PR #6, opened 2020-12-14
Verify Quarkus Http/2 test-plan

pjgg:


QUARKUS-279.md
```diff
@@ -0,0 +1,42 @@
...
+
+Server Side:
+- Verify that http/2 is set as default HTTP protocol
+- Consume and endpoint developed with `quarkus-resteasy-jsonb` through HTTP/2 Client (Vertx HTTP Client)
+
+### Impact on testsuites and testing automation:
+
+
+### Impact on resources:
+ - More services will be running on OpenShift cluster, impact on CPU/memory should be minimal
```
rsvoboda, line 25: How many modules do you expect to be added into OpenShift TS?

In other words also asking if all the above can be covered in one module. Build + deployment time is the largest time portion from the overall execution time.

Execution time implication will be lower if you are able to limit number of new modules

QUARKUS-279.md
```diff
@@ -0,0 +1,42 @@
...
+ 
+## Automated test development
+This step should ensure:
+ - Focus on high-level scenarios with multiple components and features involved
+ - OpenShift should be the primary target platform for product to ensure good interoperability
+ - Ensure platform specific features or components have appropriate test coverage
+
+## Advanced topics for test development
+In the future we should extends this test to databases that support HTTP/2 as:
+ - Elasticsearch Rest Client
```
rsvoboda, line 40: Elasticsearch topic is not a top priority even on Quarkus community side. There were some discussion around elasticsearch and some people think it's to "heavy" for Quarkus and the should be different options evaluated.

I would drop Advanced topics section for this TP.

rsvoboda (changes requested):

QUARKUS-279.md
pjgg: Yes, I think that this TP could be done in just one:

Have look this preview: https://github.com/quarkus-qe/quarkus-openshift-test-suite/pull/146

An existing module was reused (Http), I think that make sense.

rsvoboda (approved):

PR #10, opened 2021-01-29
Add test plan for Quarkus-712 Rest Client support for Mutiny

pjgg:


QUARKUS-712.md
```diff
@@ -0,0 +1,26 @@
+# Quarkus-712 - Rest Client support for Mutiny
+
+The goal of this test plan is to cover Quarkus/Mutiny rest Client. 
+
+JIRA link: https://issues.redhat.com/browse/QUARKUS-712 
+
+## Scope of the testing
+Following actions were taken to ensure test coverage and automation for Quarkus/Mutiny rest Client:
+ - Setup async endpoints secured and unsecured and cover the main HTTP methods (Get, Post, Put, Delete)
+ - Setup async HTTP client pointing to our endpoints.
```
rsvoboda, line 10:
```suggestion
 - Setup async quarkus-rest-client-mutiny based client pointing to our endpoints.
```

QUARKUS-712.md
```diff
@@ -0,0 +1,26 @@
...
+
+### Impact on testsuites and testing automation:
+ - Ensure this coverage works both in JVM and NATIVE mode
+ - Integration tests covered by beefy-scenarios testsuite
```
rsvoboda, line 14: Please check what is covered in https://github.com/quarkusio/quarkus/tree/master/integration-tests/resteasy-mutiny and make sure that the addition into beefy-scenarios testsuite covers (at least) slightly different scenario.

rsvoboda: Just for the record I think we have more complex scenario thanks to oidc involvement in that module. I just want you to double check upstream test coverage.

rsvoboda (changes requested):

QUARKUS-712.md
pjgg: Upstream is focus on Get Http method. Also cover SSE (we didn't cover this protocol).
We are adding some extra covering as Auth plust POST, PUT and DELETE methods. 

Also, our tests are the same for all HttpClients (sync/async), which is interesting because we are verifying that we give the same behavior.

rsvoboda (approved):

PR #15, opened 2021-02-03
Adding a test plan for QUARKUS-715 Support for multiple Hibernate ORM persistence units.

mjurc:


QUARKUS-715.md
```diff
@@ -0,0 +1,55 @@
+# QUARKUS-715 Support for multiple Hibernate ORM persistence units
+
+This document describes a high-level testing overview of [QUARKUS-715 Support for multiple Hibernate ORM persistence units](https://issues.redhat.com/browse/QUARKUS-715).
+
+## Scope of the testing
+
+### Component testing
```
Sgitario, line 7: This was really useful information! As Hibernate is a third party component, I would move this later in this test plan as an extra bits or something.

QUARKUS-715.md
```diff
@@ -0,0 +1,55 @@
...
+These tests are executed in Quarkus CI on upstream on a subset of supported platforms.
+
+An integration test verifying correctness of the feature on OpenShift clusters will be implemented in
+[quarkus-qe/quarkus-openshift-test-suite](https://github.com/quarkus-qe/quarkus-openshift-test-suite). The test will be
+an extension of existing SQL test cases to use application with multiple Hibernate persistence units.
+
+[quarkus-qe/beefy-scenarios `004-quarkus-HHH-and-HV`](https://github.com/quarkus-qe/beefy-scenarios/tree/master/004-quarkus-HHH-and-HV) module will also be extended with an application using multiple
+Hibernate persistence units.
+
+### Platform testing
```
Sgitario, line 24: Maybe we could have this section as a shared page for all test plans?

Sgitario (approved):
I really liked how you wrote this test plan! As I stated in some comments: 
- I would add another section to state the functional/non-functional requirements and use cases and say where these requirements are covered.
- I don't think we need the `Getting familiar with the feature` as it does not add value from my point of view. 

But this is only my opinion and I'm not confident enough about the overall process yet, so up to you and @rsvoboda to finally approve the test plan.

However, what I think is a must is that we all should write the test plans in the same way (using this approach or using others) and hence make an effort to propose a common template. Though, this is out of the scope of this test plan. @rsvoboda @mjurc what about creating a task for doing this?

QUARKUS-715.md
```diff
@@ -0,0 +1,55 @@
+# QUARKUS-715 Support for multiple Hibernate ORM persistence units
+
+This document describes a high-level testing overview of [QUARKUS-715 Support for multiple Hibernate ORM persistence units](https://issues.redhat.com/browse/QUARKUS-715).
```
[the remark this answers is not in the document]
rsvoboda: I think JIRA is accurate, you must differentiate between pure datasources and ORM

 - https://quarkus.io/blog/quarkus-1-8-0-final-released/#multiple-persistence-units (1.8 introduces support for multiple persistence units in the Hibernate ORM extension.)
 
 - https://quarkus.io/blog/quarkus-1-10-2-final-released/#more-reactive-changes (Reactive SQL Clients now support multiple datasources)

QUARKUS-715.md
rsvoboda: This is changing over time and also not applicable to all RFEs.

Michal could just go with "Tests are going to be executed against all tested versions of databases, RHEL and OpenShift versions. Both JVM and Native mode is gonna be tested."

QUARKUS-715.md
```diff
@@ -0,0 +1,55 @@
...
+## Scope of the testing
+
+### Component testing
+The feature consists of boilerplate code allowing the Quarkus applications to hold more than one Hibernate persistence 
+unit in their context.
+
+Hibernate functionality is tested on [Hibernate CI](https://ci.hibernate.org/).
+
+### Integration testing
+Basic set of integration tests has been implemented and merged with the [feature implementation](https://github.com/quarkusio/quarkus/pull/11322/files).
```
rsvoboda, line 14: Basic set => Set

QUARKUS-715.md
```diff
@@ -0,0 +1,55 @@
...
+These tests are executed in Quarkus CI on upstream on a subset of supported platforms.
+
+An integration test verifying correctness of the feature on OpenShift clusters will be implemented in
+[quarkus-qe/quarkus-openshift-test-suite](https://github.com/quarkus-qe/quarkus-openshift-test-suite). The test will be
+an extension of existing SQL test cases to use application with multiple Hibernate persistence units.
+
+[quarkus-qe/beefy-scenarios `004-quarkus-HHH-and-HV`](https://github.com/quarkus-qe/beefy-scenarios/tree/master/004-quarkus-HHH-and-HV) module will also be extended with an application using multiple
+Hibernate persistence units.
```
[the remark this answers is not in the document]
rsvoboda: I think it's still good to have coverage for multiple DBs in BeefyTS, we run it against Mandrel, upstream runs against GraalVM. 
Would go with new module tbh, combination of MariaDB + PostgreSQL + MSSQL

rsvoboda (changes requested):
Added few comment ++ I miss info about impact on QE infrastructure

QUARKUS-715.md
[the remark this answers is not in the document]
mjurc: Exactly - upstream database coverage is limited.

QUARKUS-715.md
mjurc: The issue here is we don't have a tested configurations document yet. In EAP, we had document that was updated when each new release was planned/scoped and test plans referred to that.

I've updated the section in accordance with Rosťa's suggestion :)

mjurc:
> I really liked how you wrote this test plan! As I stated in some comments:
> 
>     * I would add another section to state the functional/non-functional requirements and use cases and say where these requirements are covered.
> 
>     * I don't think we need the `Getting familiar with the feature` as it does not add value from my point of view.
> 
> 
> But this is only my opinion and I'm not confident enough about the overall process yet, so up to you and @rsvoboda to finally approve the test plan.
> 
> However, what I think is a must is that we all should write the test plans in the same way (using this approach or using others) and hence make an effort to propose a common template. Though, this is out of the scope of this test plan. @rsvoboda @mjurc what about creating a task for doing this?

I'd personally like this. Some sections are completely the same in each of the test plans, on the other hand, some of the test plans lack any sort of tracking of the 'tiers' of the feature testing, ranging from upstream CIs to our job executions.

rsvoboda (approved):

PR #22, opened 2021-05-06
Test Plan for Quarkus-874: Production support for Windows

Sgitario:


QUARKUS-874.md
```diff
@@ -0,0 +1,36 @@
...
+ - Beefy Test Suite
+   - In Github CI - for Pull Requests and Daily runs
+   - New Jenkins Jobs
+ - QuickStarts Test Suite
+   - New Jenkins Jobs
+ - Combinations of Quarkus extensions Test Suite
+   - In Github CI - for Pull Requests and Daily runs
+ - Workshop Test Suite
+   - New Jenkins Jobs
+
```
Sgitario, line 26: Do we need to update OpenShift Test Suite as well? As far I see it, the test suite is not OS dependant.

QUARKUS-874.md
```diff
@@ -0,0 +1,36 @@
...
+### Platform testing
+Tests are going to be executed against all tested versions of databases, RHEL and OpenShift versions. 
+Both the JVM and native mode is going to be tested.
+
+## Impact on resources
+- New Windows instances in our Jenkins environments
+  - Windows instance must support Native compilation (GraalVM, VS Build Tools) and Docker (Windows Containers)
+
```
Sgitario, line 34: Should we state the existing limitation around testcontainers / Linux containers on Windows? I think this will be eventually addressed, so we don't need to state it here.

QUARKUS-874.md
```diff
@@ -0,0 +1,37 @@
+# QUARKUS-874 Production support for Windows
+
+JIRA link: https://issues.redhat.com/browse/QUARKUS-874
+
+The goal is to ensure some level of support and identify existing limitations on Quarkus side for Windows.
+
+## Scope of the testing
+The selected Test Suites will be executed on the selected Windows Platforms to ensure functionality is fully accomplished.
+In order to accommodate this in our current test suites, we need to rename existing jobs in Jenkins in order to differentiate them for those running on RHEL and on Windows platforms. 
+
```
rsvoboda, line 10: Pls mention that some testsuites will need adjustments for correct execution on Windows.

QUARKUS-874.md
```diff
@@ -0,0 +1,37 @@
...
+   - In Github CI - for Pull Requests
+   - New Jenkins Jobs
+ - Beefy Test Suite
+   - In Github CI - for Pull Requests and Daily runs
+   - New Jenkins Jobs
+ - QuickStarts Test Suite
+   - New Jenkins Jobs
+ - Combinations of Quarkus extensions Test Suite
+   - In Github CI - for Pull Requests and Daily runs
+ - Workshop Test Suite
```
rsvoboda, line 25: I think this TS is no longer used. @mjurc can you clarify?

QUARKUS-874.md
rsvoboda: For OpenShift Test Suite we should run some selected modules like the deployment strategies to ensure some functionality baseline checks.

QUARKUS-874.md
```diff
@@ -0,0 +1,37 @@
...
+   - New Jenkins Jobs
+
+### Platform testing
+Tests are going to be executed against all tested versions of databases, RHEL and OpenShift versions.
```
rsvoboda, line 29: This is not true, we will need to be cautious about the expansion of the executed jobs.

Use something like this:
```
JVM mode on is the focus for Windows platform. This RFE does not include native executable (Mandrel) or any extension that requires native libraries (like Kafka streams client).
```

QUARKUS-874.md
```diff
@@ -0,0 +1,37 @@
...
+ - Workshop Test Suite
+   - New Jenkins Jobs
+
+### Platform testing
+Tests are going to be executed against all tested versions of databases, RHEL and OpenShift versions. 
+Both the JVM and native mode is going to be tested.
+
+## Impact on resources
+- New Windows instances in our Jenkins environments
+  - Windows instance must support Native compilation (GraalVM, VS Build Tools) and Docker (Windows Containers)
```
rsvoboda, line 34: Native compilation (GraalVM, VS Build Tools) is not required by this RFE, I would drop it from this text.
I want that to be available so we have option to try native binaries on Windows eventhough it's not this RFE requirement.

Docker (Windows Containers) is not what we want. We need Docker with Linux containers support.
Our testsuites rely on testcontainers to assure portability between different CI environments and to avoid dependency on many external services.

I miss impact on PSI capacity and execution time of our tests.
 - Windows instances will be executed next to the RHEL instances => more CPU / memory in the lab would be needed to ensure similar execution time as now.
 - We do not have new capacity in the lab => test execution will be prolonged (doubled in the worse case)
 - Windows virtual machines usually require more resources than RHEL ones, expecting `large` or `xlarge` flavor to be used
 - More jobs pulling artifacts from Artifactory instance => we may need to use more resources for Artifactory

QUARKUS-874.md
rsvoboda: I would mention that, it causes us a lot of troubles. And the available solution doesn't work out of the box / we (Tomas mainly) spent a lot of hours on this topic.

QUARKUS-874.md
```diff
@@ -0,0 +1,37 @@
...
+ - Combinations of Quarkus extensions Test Suite
+   - In Github CI - for Pull Requests and Daily runs
+ - Workshop Test Suite
+   - New Jenkins Jobs
+
+### Platform testing
+Tests are going to be executed against all tested versions of databases, RHEL and OpenShift versions. 
+Both the JVM and native mode is going to be tested.
+
+## Impact on resources
```
rsvoboda, line 32: Add "Impact on Jenkins" section to mention need of new jobs definitions and implication on some jobs naming unification

QUARKUS-874.md
```diff
@@ -0,0 +1,37 @@
...
+JIRA link: https://issues.redhat.com/browse/QUARKUS-874
+
+The goal is to ensure some level of support and identify existing limitations on Quarkus side for Windows.
+
+## Scope of the testing
+The selected Test Suites will be executed on the selected Windows Platforms to ensure functionality is fully accomplished.
+In order to accommodate this in our current test suites, we need to rename existing jobs in Jenkins in order to differentiate them for those running on RHEL and on Windows platforms. 
+
+### Selected Windows Platforms
+- Windows Server 2019
```
rsvoboda, line 12: Please add a note that we will make sure to have access to Windows 10 machine in PSI when needed in case of fix for customer issue bound to that particular environment.

QUARKUS-874.md
```diff
@@ -0,0 +1,37 @@
...
+
+### Selected Test Suites
+ - Start-Stop Test Suite
+   - In Github CI - for Pull Requests
+   - New Jenkins Jobs
+ - Beefy Test Suite
+   - In Github CI - for Pull Requests and Daily runs
+   - New Jenkins Jobs
+ - QuickStarts Test Suite
+   - New Jenkins Jobs
```
rsvoboda, line 22: I'm thinking if we need both QS and Beefy TS jobs for Windows.
I think we could just have Beefy TS.

rsvoboda (changes requested):

QUARKUS-874.md
mjurc: We still use it as simple test on bare metal

QUARKUS-874.md
```diff
@@ -0,0 +1,37 @@
+# QUARKUS-874 Production support for Windows
+
+JIRA link: https://issues.redhat.com/browse/QUARKUS-874
+
+The goal is to ensure some level of support and identify existing limitations on Quarkus side for Windows.
```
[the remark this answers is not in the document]
Sgitario: Done

QUARKUS-874.md
Sgitario: Added

QUARKUS-874.md
Sgitario: Added

QUARKUS-874.md
Sgitario: Then I will keep it here.

QUARKUS-874.md
Sgitario: Ok, removed

QUARKUS-874.md
Sgitario: Good point. Added.

QUARKUS-874.md
Sgitario: This section was a leftover, thanks for spoting. I've already updated it using your suggestion.

QUARKUS-874.md
rsvoboda: Question is if we need such TS on Windows.
For me BeefyTS coverage is enough, Workshop job just for RHEL to ensure functionality.
Or is there anything specific / risky for Windows why Workshop would be needed?

QUARKUS-874.md
Sgitario: Added as part of the Impact on resources section and I provided also additional context.

QUARKUS-874.md
Sgitario: Fixed

QUARKUS-874.md
Sgitario: Done

Sgitario:
@rsvoboda I think I've already addressed all your comments, would you mind to take another look?

QUARKUS-874.md
```diff
@@ -0,0 +1,50 @@
...
+ - Rename existing jobs to differentiate them for those running on RHEL and on Windows platforms
+
+## Impact on resources
+The major impact is on the PSI and Jenkins capacity:
+ - Windows instances will be executed next to the RHEL instances 
+   - Overall test execution for RHEL and Windows jobs will be affected (doubled in the worse case)
+   - Windows virtual machines usually require more resources than RHEL ones, expecting large or xlarge flavor to be used
+   - More CPU / memory in the lab would be needed to ensure similar execution time as now.
+ - Remote Docker machines instances based on RHEL to build container images by Windows jobs
+   - This is required by the Windows jobs for some test suites that require third parties components such us databases and these components are based on Linux images
```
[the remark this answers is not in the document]
Sgitario: Removed

Sgitario:
@rsvoboda I've incorporated all your commit suggestions and dropped the line you requested.
Would you mind to take another look? From my point of view, the only remaining point is if we should cover Workshop TS in Windows or not. I didn't add it for now.

rsvoboda (approved):

rsvoboda:
I plan to give @mjurc one more day, otherwise merging on Friday 2 PM Brno time.

PR #23, opened 2021-05-06
Test plan for QUARKUS-508

mjurc:
* Adding a test plan for QUARKUS-508 - Create one or more quarkus devfile v2,
  ensure odo based workflow


rsvoboda (changes requested):
There is no content in the file 🙃

mjurc:
@rsvoboda Sorry about that, it should be fixed now

QUARKUS-508.md
```diff
@@ -0,0 +1,35 @@
...
+- Define the scope of the testing
+- Analyze the possible impact on Quarkus itself - e.g performance, compatibility
+- Determine impact on test suites - e.g. new test suite, increased complexity of existing test suite, maintenance burden
+- Determine impact on QE infrastructure - e.g. increased complexity of existing automation, maintenance burden, need for new external resources
+
+### Impact on testsuites and testing automation:
+A deployment strategy utilising `odo` tool will be implemented in [`quarkus-test-framework`](https://github.com/quarkus-qe/quarkus-test-framework/).
+A test using the `odo` deployment strategy will deploy a simple REST endpoint and verify that the endpoint responds.
+This test will be implemented in [`quarkus-test-suite`](https://github.com/quarkus-qe/quarkus-test-suite/) project.
+
```
rsvoboda, line 19: Quarkus tests should be executed when `odo` tool performs new release too.
We should start discussion with the counter parts, please ping me to give you some names

rsvoboda: And also mention this in the test plan

rsvoboda (changes requested):

QUARKUS-508.md
mjurc: I am not sure I agree on this part, we don't test new releases of `oc` for example.

`odo`, on the RH side, is provided as part of OpenShift dev tools, tied to OCP release.

QUARKUS-508.md
```diff
@@ -0,0 +1,35 @@
...
+### Impact on resources:
+The new test implemented in Quarkus OpenShift test suite does not require infrastructural changes and its execution
+should be finished within minutes. Test will be executed on all the tested OCP platforms.
+
```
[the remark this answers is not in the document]
mjurc: Updated.

QUARKUS-508.md
```diff
@@ -0,0 +1,35 @@
...
+- Determine impact on QE infrastructure - e.g. increased complexity of existing automation, maintenance burden, need for new external resources
+
+### Impact on testsuites and testing automation:
+A deployment strategy utilising `odo` tool will be implemented in [`quarkus-test-framework`](https://github.com/quarkus-qe/quarkus-test-framework/).
+A test using the `odo` deployment strategy will deploy a simple REST endpoint and verify that the endpoint responds.
+This test will be implemented in [`quarkus-test-suite`](https://github.com/quarkus-qe/quarkus-test-suite/) project.
+
+### Impact on resources:
+The new test implemented in Quarkus OpenShift test suite does not require infrastructural changes and its execution
+should be finished within minutes. Test will be executed on all the tested OCP platforms.
```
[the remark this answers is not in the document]
mjurc: Technically, I agree - there shouldn't be any issues as `odo` fills out kubernetes object templates and then executes the commands defined in devfiles.

On the other hand, we're not testing `odo`, we're testing the Quarkus devfile.

QUARKUS-508.md
rsvoboda: We will test when we have the release. The point if this note is that the odo team should test scenario with Quarkus when they do the release of odo to early notice possible issues.
We should at least start discussion as part of this RFE.

QUARKUS-508.md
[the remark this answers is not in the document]
rsvoboda: We can't ensure every combination, we need to start running on one main OCP platform and define our expectations.

It should be detected by devfiles team / odo team when devfile works on one OCP version and not on another supported OCP version. This is not subject of our testing.

QUARKUS-508.md
```diff
@@ -0,0 +1,46 @@
...
+test will be ran in a separate, new job as a part of post-GA tasks, verifying that the `redhat-product` starter 
+application defined in the `devfile` produces the correct productised starter application.
+
+A deployment strategy utilising `odo` tool will be implemented in [`quarkus-test-framework`](https://github.com/quarkus-qe/quarkus-test-framework/).
+
+A test using the `odo` deployment strategy will deploy a simple REST endpoint and verify that the endpoint responds.
+This test will be implemented in [`quarkus-test-suite`](https://github.com/quarkus-qe/quarkus-test-suite/) project.
+
+### Impact on resources:
+Jenkins workers will need to be updated with `odo` tool.
```
rsvoboda, line 26: OC tool is downloaded as part of the job execution. Why do we want to have `odo` as standard part of worker node. It can get old over time if the image is not regularly updated.

QUARKUS-508.md
```diff
@@ -0,0 +1,46 @@
...
+
+The new job testing the starter application defined in the `java-quarkus` component devfile will be implemented and ran 
+in the standalone, as part of post-GA tasks. This execution should be completed within minutes.
```
rsvoboda, line 29: So there is no possibility to change devfile repository to allow testing before GA

rsvoboda (changes requested):

QUARKUS-508.md
mjurc: Well, there's two options:

- download as part of job execution
- install by yum

I'm not really particular to either

QUARKUS-508.md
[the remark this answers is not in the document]
mjurc: Right, the RFE is on our side so I assumed we owned the devfile. 

I will check with odo team if they test the default registry devfiles or if there's any unified owner of the repository.

As for every combination... Right, odo 2 should work with every version of OCP4. My concern is not about the kubernetes resources that odo generates, but about the commands we have in Quarkus dev file. For example https://github.com/odo-devfiles/registry/blob/master/devfiles/java-quarkus/devfile.yaml#L32

QUARKUS-508.md
mjurc: Actually, I was wrong - let me update again :slightly_smiling_face:

rsvoboda (approved):

PR #26, opened 2021-06-30
Plans for Hibernate Reactive tests

fedinskiy:
Test plans for Hibernate Reactive and Hibernate Reactive Panache modules, both (as of v2.0.0)currently in Tech Preview status. Part of wider https://issues.redhat.com/browse/QUARKUS-962 and https://issues.redhat.com/browse/QUARKUS-1079 issues

### DIFF
--- a/QUARKUS-1079.md
+++ b/QUARKUS-1079.md
@@ -1,0 +1,74 @@
+# QUARKUS-1079 - Hibernate Reactive as Tech Preview
+
+JIRA link: https://issues.redhat.com/browse/QUARKUS-1079
+
+For native mode similar test coverage as for JVM mode needs to be ensured.
+
+## Scope of the testing
+This feature added previews of two extensions:
+ - Hibernate Reactive
+ - Hibernate Reactive with Panache
+
+Following test coverage should be implemented for Hibernate Reactive:
+- Usage of application.properties file for passing reactive-hibernate settings
+- Usage of import.sql
+- Smoke test(getting a working Mutiny session)
+- Retrieving Multi result from the DB
+- Retrieving Uni result(require a transformation) from the DB
+- Usage of DB-generated ids and ids generated with Java code(via ReactiveIdentifierGenerator)
+- Validation of values
+- Attribute converters
+- Usage of session opened through openSession() method as well as withSession()
+- Usage of transaction started through openSession() method as well as withSession()
+- Usage of session and transaction opened with the same method
+- Session methods: find, persist, remove, merge, flush
+- Queries: regular, native, named, JPA criteria
+- Query methods: setParameter, setMaxResults, getFirstResult, getSingleResult, getResultList, executeUpdate
+- Eager and lazy fetches
+- Field-level lazy fetching(Hibernate team "doesn't encourage" using this feature, though)
+- Stateless sessions and their methods
+
+Following test coverage should be implemented for Hibernate Reactive with Panache:
+- Active record pattern
+- Repository pattern
+- list and stream methods
+- find method
+- sorting
+- Named queries
+- DTO projection
+- Transactions
+- Id generation
+
+All tests should run for PostgreSQL, MySQL, DB2, MS SQL and Oracle SQL.
+
+### Impact on testsuites and testing automation:
+ - Ensure this coverage works both in JVM and NATIVE mode
+ - Some smoke tests were created in "examples" section of quarkus-test-framework, they should be moved into quarkus-test-suite and all subsequent tests should be created there.
+
+### Impact on resources:
+ - More services will be running on OpenShift cluster, impact on CPU/memory should be minimal
+ - Bare metal tests will deploy docker containers with DBs, but they are mostly lightweight
+ - Only exception is a DB2 database, which consumes a lot of resources, so there should be an option to disable running tests for it.
+ - Test execution will be prolonged, preliminarily estimation is ~10 minutes per each OCP stream
+ - QUARKUS-1079 covers both JVM and NATIVE mode, native image imposes multiple times prolonged execution time and increased memory requirements for the build
+
+Our current testing capacity is sufficient to cover this testing, execution time will be increased though.
+
+If multiple streams get introduced we need to have increased capacity in the lab.
+
+## Getting familiar with the feature
+Following actions were taken to ensure familiarity:
+ - Focus on exploratory testing of the features
+ - Ensure good user experience and simplicity of use
+
+## Automated test development
+This step should ensure:
+ - Focus on high-level scenarios with multiple components and features involved
+ - OpenShift should be the primary target platform for product to ensure good interoperability
+ - Ensure platform specific features or components have appropriate test coverage
+
+## Advanced topics
+    - Locking
+    - Multitenancy
+    - Cache
+    - Tuning and perfomance


QUARKUS-1079.md
```diff
@@ -0,0 +1,73 @@
...
+- Repository pattern
+- list and stream methods
+- find method
+- sorting 
+- Named queries
+- DTO projection
+- Transactions
+- Id generation
+
+Ideally, all of aforementioned tests should run for PostgreSQL, MySQL and DB2 connections.
```
mjurc, line 43: What about the other supported databases, e.g. MSSQL and Maria?

QUARKUS-1079.md
```diff
@@ -0,0 +1,73 @@
...
+Following test coverage should be implemented for Hibernate Reactive with Panache:
+- Active record pattern
+- Repository pattern
+- list and stream methods
+- find method
+- sorting 
+- Named queries
+- DTO projection
+- Transactions
+- Id generation
```
mjurc, line 41: I think we already cover some of these scenarios. Can you maybe list the scenarios we already have implemented and scenarios that will need to be implemented?

QUARKUS-1079.md
```diff
@@ -0,0 +1,73 @@
...
+
+Ideally, all of aforementioned tests should run for PostgreSQL, MySQL and DB2 connections.
+
+### Impact on testsuites and testing automation:
+ - Ensure this coverage works both in JVM and NATIVE mode
+ - Tests were created in "examples" section of quarkus-test-framework, maybe they can be moved to quarkus-test-suite
```
mjurc, line 47: `examples` module in `quarkus-test-framework` are just examples for implementation.

We should definitely move them to `quarkus-test-suite` :smile:

QUARKUS-1079.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Impact on resources:
+ - More services will be running on OpenShift cluster, impact on CPU/memory should be minimal
```
mjurc, line 50: We should probably run a (sub)set of the scenarios on bare metal as well, with test containers for DB provisioning

QUARKUS-1079.md
```diff
@@ -0,0 +1,73 @@
...
+- Transactions
+- Id generation
+
+Ideally, all of aforementioned tests should run for PostgreSQL, MySQL and DB2 connections.
+
+### Impact on testsuites and testing automation:
+ - Ensure this coverage works both in JVM and NATIVE mode
+ - Tests were created in "examples" section of quarkus-test-framework, maybe they can be moved to quarkus-test-suite
+
+### Impact on resources:
```
mjurc, line 49: Do we have any general idea how long the tests would run? This should perhaps be mentioned in the test plan as well.

mjurc (changes requested):

QUARKUS-1079.md
fedinskiy: I based my assumptions on this phrase: "Out of the box, the Vert.x clients for PostgreSQL, MySQL, and DB2 are supported"[1]. Is there a better source?

[1] http://hibernate.org/reactive/documentation/1.0/reference/html_single/#preface

QUARKUS-1079.md
fedinskiy: It is: 
https://github.com/quarkus-qe/quarkus-test-plans/pull/26/files/86af3fa10225a22958266178ae6d47bd966d1c2f#diff-a5aa5d754f3f0716f560b5649d940c4220be1ff3baa4889dd4ed59686872f3c9R51

QUARKUS-1079.md
fedinskiy: Done

QUARKUS-1079.md
fedinskiy: Ok. I changed "maybe" to "should"

QUARKUS-1079.md
fedinskiy: I added considerations for bare metal tests.

fedinskiy:
@kshpak could you take a look?

rsvoboda:
Please update the PR to comply with https://gitlab.cee.redhat.com/quarkus-qe/signpost/-/blob/main/pull-requests.md#main-rule-set-for-pull-requests

fedinskiy:
@rsvoboda changed name and description to better reflect the changes in the PR.

QUARKUS-1079.md
```diff
@@ -0,0 +1,74 @@
...
+
+For native mode similar test coverage as for JVM mode needs to be ensured.
+
+## Scope of the testing
+This feature added previews of two extensions: 
+ - Hibernate Reactive
+ - Hibernate Reactive with Panache
+
+Following test coverage should be implemented for Hibernate Reactive:
+- Usage of application.properties file for passing reactive-hibernate settings [examples]
```
rsvoboda, line 13: Why is `[examples]` text in your Test Plan items?

QUARKUS-1079.md
```diff
@@ -0,0 +1,74 @@
...
+- Named queries
+- DTO projection
+- Transactions
+- Id generation
+
+Ideally, all of aforementioned tests should run for PostgreSQL[examples], MySQL and DB2 connections.
+
+### Impact on testsuites and testing automation:
+ - Ensure this coverage works both in JVM and NATIVE mode
+ - Some smoke tests were created in "examples" section of quarkus-test-framework, more advanced should be put into quarkus-test-suite
```
rsvoboda, line 47: Test coverage should be mainly in https://github.com/quarkus-qe/quarkus-test-suite/

Examples directory should contain just one sample, it's not the main place for test coverage.

QUARKUS-1079.md
```diff
@@ -0,0 +1,74 @@
...
+- Repository pattern
+- list and stream methods
+- find method
+- sorting 
+- Named queries
+- DTO projection
+- Transactions
+- Id generation
+
+Ideally, all of aforementioned tests should run for PostgreSQL[examples], MySQL and DB2 connections.
```
rsvoboda, line 43: There are quite a lot of items listed above.

Do you have estimate how long it would take to create such coverage?

Did you check what's covered under https://github.com/quarkusio/quarkus/tree/main/integration-tests ?

QUARKUS-1079.md
```diff
@@ -0,0 +1,74 @@
...
+ - Ensure this coverage works both in JVM and NATIVE mode
+ - Some smoke tests were created in "examples" section of quarkus-test-framework, more advanced should be put into quarkus-test-suite
+
+### Impact on resources:
+ - More services will be running on OpenShift cluster, impact on CPU/memory should be minimal
+ - Bare metal tests will deploy docker containers with DBs, but they are mostly lightweight
+ - Test execution will be prolonged, preliminarily estimation is ~10 minutes per each OCP stream
+ - QUARKUS-1079 covers both JVM and NATIVE mode, Native image imposes multiple times prolonged execution time and increased memory requirements for the build
+
+Our current testing capacity is sufficient to cover this testing.
```
rsvoboda, line 55:
```suggestion
Our current testing capacity is sufficient to cover this testing, execution time will be increased though.
```

QUARKUS-1079.md
```diff
@@ -0,0 +1,74 @@
...
+
+Ideally, all of aforementioned tests should run for PostgreSQL[examples], MySQL and DB2 connections.
+
+### Impact on testsuites and testing automation:
+ - Ensure this coverage works both in JVM and NATIVE mode
+ - Some smoke tests were created in "examples" section of quarkus-test-framework, more advanced should be put into quarkus-test-suite
+
+### Impact on resources:
+ - More services will be running on OpenShift cluster, impact on CPU/memory should be minimal
+ - Bare metal tests will deploy docker containers with DBs, but they are mostly lightweight
```
rsvoboda, line 51: DB2 for example takes minutes to start and is heavy on memory

rsvoboda (changes requested):
For sure check what is covered in https://github.com/quarkusio/quarkus/tree/main/integration-tests 

Some of the test coverage enhancements can be added into upstream TS - https://github.com/quarkusio/quarkus/tree/main/integration-tests 

You should also consider creating something like https://gitlab.cee.redhat.com/quarkus-qe/jenkins-jobs/-/blob/main/jobs/quarkus/main_rhel8_jdk11_database_ts.groovy for reactive to ensure different versions of database with upstream based tests.

QUARKUS-1079.md
fedinskiy: It describes, which test suite contains these tests. It's either [examples], which means "examples" section in framework, or [testsuite], which means "new" test suite

QUARKUS-1079.md
fedinskiy: Do we spin db2 on Openshift directly? In Vertx we were using dballocator for this purpose.

rsvoboda: We do not use dballocator to avoid hard dependency on this service, DBs are started in container to have better portability and easier local investigation if case fails.

rsvoboda:
@fedinskiy when do you plan to have update of this TP ?

QUARKUS-1079.md
fedinskiy: @rsvoboda I have two questions:
1) Which databases should we cover? Is there is a list or something?
2) Are any of them comparable to db2 in terms of speed and resources used? I checked Postgres, Maria, My  and MS SQL, and they seem to be relatively fast. Brief search for resource consumption showed it to be below 2 GBs, which is half of what is required for Docker, so it should not be a problem.

QUARKUS-1079.md
fedinskiy: Most of them are trivial, so I would say between two or three human-weeks.

Now I did. Which tests should go into integration tests and which into our new test suite? As I can see, our ":new" TS supports OpenShift deployment and it's easier to support new databases in there. Are there any other benefits?

QUARKUS-1079.md
fedinskiy: Changed this section

QUARKUS-1079.md
rsvoboda: quarkus/tree/main/integration-tests ... the main idea is that we should not duplicate test coverage for existing coverage. I was assuming that some parts of your proposal can be already implemented and executed in upstream as part of every PR in both JVM and Native mode.

Areas that are not covered are fine to be in our new TS.

QUARKUS-1079.md
rsvoboda: Details are in the umbrella epic in JIRA - https://issues.redhat.com/browse/QUARKUS-962
Thomas wrote:
 -  PostgreSQL and MySQL is required
 -  DB2, MSSQL and Oracle as tech preview is important but not required

MS SQL reactive client was added in https://github.com/quarkusio/quarkus/releases/tag/2.1.0.CR1
From my side we need to cover also TP databases as they will be promoted to fully supported at some time.

DB2 is probably the worse from CPU/RAM perspective.

QUARKUS-1079.md
fedinskiy: As far as I can see, most(all?) areas are covered in integration tests. The thing is, there is a lack of coverage in terms of databases(panache is tested only with postgres, there is no coverage for oracle and mssql at all). Also, as far as I understand, they do not test anything on OpenShift(or Kubernetes).

So, I suggest we put these thing into our TS:
1) Basic set-up operations(app.properties, sql-load-script)
2) Basic operations for every database in(find, persist, remove, merge,flush)
3) Advanced things, if we suspect, that they may differ between databases(queries, id generation, transactions)
All of aforementioned should be implemented either for pure reactive Hibernate or for Panache, depends in which one is more popular among RH clients.
On the other hand, we should implement these things in integration tests:
1) Attribute converters
2) Tests currently implemented for Hibernate, if we will implement only Panache tests or vice versa.

What do you think?

rsvoboda: > there is a lack of coverage in terms of databases(panache is tested only with postgres, there is no coverage for oracle and mssql at all).

imho we should contribute upstream to enhance the set, at least for mysql/mariadb and mssql.

> they do not test anything on OpenShift(or Kubernetes).

That's true, on OpenShift we do not need to cover every scenario tested in integration-tests
Upstream is working on introducing OpenShift an Kubernetes tests as part of the CI.

About our TS, I would start with pure Reactive Hibernate, reactive support for Panache is quite new.

fedinskiy: The biggest problem of integration-tests — it's a zoo. There are four maven modules(postgres, mysql, db2, panache) and each one contains different methods and entity classes, which duplicate each other. In order to check a behaviour for different db's, we should create new test at least three times and each time it should work with a different data class.
As alternative, we can create a separate "unified" module, but I'm afraid it will only worsen the situation[1]

It's interesting, since Reactive Panache looks more polished(I found that field validation works in Panache HR, but doesn't work in Standard HR) and guide for it is at least in the works[2](no Hibernate Reactive guide exists at the moment).

[1] https://xkcd.com/927/
[2] https://github.com/quarkusio/quarkus/issues/18103

QUARKUS-1079.md
fedinskiy: Added info about DB2

QUARKUS-1079.md
rsvoboda: so this sounds like we skip integration-testsuite for now and focus just on test development in our TS
about docs ... ping Sanne on the issue about the non-panache guide and plans about it
we can revisit integration-tests situation once Hibernate Reactive moves towards supported level

Please adjust the TP to say that we focus on test development in our repo

rsvoboda (changes requested):
Please drop `[implemented]` sufixes

kshpak:
Fix the commit names and squash pls

mjurc (approved):

rsvoboda (approved):

PR #33, opened 2021-08-04
Add test plan for grpc devui

kshpak:
Please get through the test plan. 
I'm not sure if this feature is suitable for the test automation since it is mostly in browser. I proposed to test it with the websocket client, but I didn't do any POC for it.


kshpak:
Since @jsmrcka is on PTO, adding @rsvoboda as a reviewer

QUARKUS-1094.md
```diff
@@ -0,0 +1,24 @@
+# QUARKUS-1094 - Initial support of the gRPC in the dev ui
+
+JIRA link: https://issues.redhat.com/browse/QUARKUS-1094
+
+Quarkus has an experimental feature for gRPC in the DevUI. Now it is possible to test gRPC services provided by the server without implementing clients. The gRPC uses HTTP2 which is well suited for microservices inner-communication, but there is no native way for browser client to utilize HTTP/2, thus the feature utilizes Web Socket Bridge for communication.
```
Sgitario, line 5: I don't understand the last sentence `there is no native way for browser client to utilize HTTP/2...`.

Sgitario (commented):
In the test framework, we use `HTMLUnit` framework to assert Dev UI. 
So, new functionality should be added into `DevModeQuarkusService` or maybe extend from it to have a `GrpcDevModeQuarkusService`.
Btw, I tried with `999-SNAPSHOT` and I could see the gRPC services, but when using it, nothing worked... any ideas? Maybe there is something not merged yet?

QUARKUS-1094.md
kshpak: Even though browsers have support for HTTP2, the browser JS doesn't have an API to work with binary data frames (because everything in HTTP2 is binary). So there are multiple solutions, the first one to use gRPC proxy (https://grpc.io/blog/state-of-grpc-web/#the-tech) and the second one, which DevUI is using, is utilizing WebSockets. The gRPC request/response can be transcoded to WebSockets (according to https://cloud.redhat.com/blog/grpc-anywhere), but WebSockets utilizing HTTP/1.1. Under the hood when using DevUI with gRPC there is a connection upgrade to `Upgrade: websocket`, because of that I don't think `htmlunit` can handle the test. 
From the above, I don't see an automation solution on HTTP level, and because the feature implements WebSockets on the server-side, the option for the test could be**:** WebSockets client that connects to WebSockets server and tests proper transcoding from gRPC to WebSockets.

P.S that's how I understood the problem, you can challenge this opinion of course

kshpak:
> In the test framework, we use `HTMLUnit` framework to assert Dev UI.
> So, new functionality should be added into `DevModeQuarkusService` or maybe extend from it to have a `GrpcDevModeQuarkusService`.
> Btw, I tried with `999-SNAPSHOT` and I could see the gRPC services, but when using it, nothing worked... any ideas? Maybe there is something not merged yet?

From quarkus main the artifact `io.quarkus.platform:quarkus-maven-plugin:999-SNAPSHOT` was not created in local repo, in fact under /platform no artifact were created.
Quarkus 2.1.0.Final works fine with discovering and testing gRPC services, some problems can be related to https://github.com/quarkusio/quarkus/issues/19218 (item 2.)

QUARKUS-1094.md
rsvoboda: Please add details from this comment (gRPC proxy + websockets way) into TP to provide additional context. 

About `htmlunit` and HTTP upgrade .. are you saying that you may not be able to automate the testing using current tool chain?

QUARKUS-1094.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Scope of the testing
+- Verify all gRPC services are listed in the DevUI
+- Verify all gRPC services are properly identified by their type
+- Verify Unary, Server Streaming, Client Streaming, Bidirectional Streaming are working properly
+- Utilize Quarkus Web Socket Client for testing (since the feature switching communication protocols from HTTP/1.1 and websocket)
+
+
+### Impact on testsuites and testing automation:
+- Test will be implemented in https://github.com/quarkus-qe/quarkus-test-suite using [quarkus-test-framework](https://github.com/quarkus-qe/quarkus-test-framework)
```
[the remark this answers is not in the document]
rsvoboda: +1

QUARKUS-1094.md
```diff
@@ -0,0 +1,24 @@
+# QUARKUS-1094 - Initial support of the gRPC in the dev ui
+
+JIRA link: https://issues.redhat.com/browse/QUARKUS-1094
+
+Quarkus has an experimental feature for gRPC in the DevUI. Now it is possible to test gRPC services provided by the server without implementing clients. The gRPC uses HTTP2 which is well suited for microservices inner-communication, but there is no native way for browser client to utilize HTTP/2, thus the feature utilizes Web Socket Bridge for communication.
+
+## Scope of the testing
+- Verify all gRPC services are listed in the DevUI
+- Verify all gRPC services are properly identified by their type
+- Verify Unary, Server Streaming, Client Streaming, Bidirectional Streaming are working properly
```
rsvoboda, line 10: Will you reuse existing apps to check the functionality of Dev UI or you need to implement the services first?

QUARKUS-1094.md
kshpak: I found some `htmlunit support for websockets`, but this should be verified. There are no really guides about that. https://sourceforge.net/p/htmlunit/code/13880/tree/trunk/htmlunit/src/test/java/com/gargoylesoftware/htmlunit/javascript/host/WebSocketTest.java#l484

QUARKUS-1094.md
kshpak: New app with required services (uni, bi-directional streaming). There is no currently TC using bi-directional streaming. Later this app can be expanded as a skeleton to build a more complex scenario

QUARKUS-1094.md
kshpak: HtmlUnit can not handle javascript execution because of its JS engine. TD implementation will be based on OkHttp library with its support for websockets

kshpak:
@rsvoboda @Sgitario test plan updated

Sgitario (approved):

rsvoboda (approved):

PR #65, opened 2022-03-04
QUARKUS-1402 Hibernate reactive test preview

fedinskiy:


QUARKUS-1402.md
```diff
@@ -41,7 +42,7 @@ Following test coverage should be implemented for Hibernate Reactive with Panach
 - Transactions
 - Id generation
 
-All tests should run for PostgreSQL, MySQL, DB2, MS SQL and Oracle SQL.
+All tests should run for PostgreSQL, MySQL, DB2, MS SQL and Oracle SQL. Oracle and DB2 are not officially supported, so failures there should not be blockers.
```
rsvoboda, line 45: Oracle and DB2 - should we cover them at all?

We do not have coverage for Oracle so fine here, DB2 is present - https://github.com/quarkus-qe/quarkus-test-suite/blob/main/sql-db/hibernate-reactive/pom.xml#L34

Should we keep it there or disable DB2 for now as it is not intended as supported option?
(would save some execution time)

rsvoboda (approved):

PR #68, opened 2022-03-14
Update Quarkus-1223 / Hibernate Search product support

pjgg:
- Reactive Resteasy test development was removed from the test plan, because currently there is an open issue that need to be fixed.
- Outbox polling is already well covered in upstream
- Postgresql and Mysql was selected as supported databases in order to verify schemas and databases multi-tenant implementation support

### DIFF
--- a/QUARKUS-1223.md
+++ b/QUARKUS-1223.md
@@ -56,11 +56,15 @@ What is missing:
 ## Automated test development
 - Enable relevant Quickstart module in Quickstart acceptance job: https://github.com/quarkusio/quarkus-quickstarts/tree/main/hibernate-search-orm-elasticsearch-quickstart.
 - Implement new test scenario in QE test suite (include OpenShift tests) combining:
-  - Multiple persistence units, outbox polling and multi-tenancy (separate schema).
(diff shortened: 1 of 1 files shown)


QUARKUS-1223.md
```diff
@@ -56,11 +56,14 @@ What is missing:
 ## Automated test development
 - Enable relevant Quickstart module in Quickstart acceptance job: https://github.com/quarkusio/quarkus-quickstarts/tree/main/hibernate-search-orm-elasticsearch-quickstart.
 - Implement new test scenario in QE test suite (include OpenShift tests) combining:
-  - Multiple persistence units, outbox polling and multi-tenancy (separate schema).
-  - RESTEasy Reactive.
```
rsvoboda, line 60: I would keep this line, add link to the blocker issue pls.

TestPlan is imho valid for this item, the fact that there is blocking issue is orthogonal to the plan.

rsvoboda (changes requested):

rsvoboda (changes requested):

rsvoboda (approved):

PR #78, opened 2022-04-19
Test plan for Temurin certification

mjurc:
* Adding a test plan for QUARKUS-1936.

### DIFF
--- a/QUARKUS-1936.md
+++ b/QUARKUS-1936.md
@@ -1,0 +1,29 @@
+# QUARKUS-1936 Certify RHBQ on Eclipse Adoptium Builds
+
+This document is a high level test plan for certification of Red Hat build of Quarkus on Temurin JVM.
+
+## Scope of the testing
+
+### Impact on test suites and testing automation
+* Temurin JVM 11 and 17 need to be added as JDKs to Quarkus QE Jenkins
+  * Temurin JVM 11 and 17 seem to be updated timely on `/qa/tools` mount point, so that will be used for the Elektra
+    time frame.
+* Temurin JVM 11 and 17 have to be added as JVMs to Polarion
+* Bare metal jobs in extended platform testing need to be extended to be ran with both Temurin 11 and 17.
+  * This affects JVM mode only.
+
+### Impact on resources
+* 1 large executor for bare metal testing for 4 hours on Temurin 11
+* 1 large executor for bare metal testing for 4 hours on Temurin 17
+
+## Future considerations
+* If Temurin JVM is not updated in timely manner on `/qa/tools` mountpoint, we may have to come up with setting up the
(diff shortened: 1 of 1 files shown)


QUARKUS-1936.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Scope of the testing
+
+### Impact on test suites and testing automation
+* Temurin JVM 11 and 17 need to be added as JDKs to Quarkus QE Jenkins
+  * Temurin JVM 11 and 17 seem to be updated timely on `/qa/tools` mount point, so that will be used for the Elektra
+    time frame.
+* Temurin JVM 11 and 17 have to be added as JVMs to Polarion
+* Bare metal and OpenShift jobs in extended platform testing need to be extended to be ran with both Temurin 11 and 17.
+* This affects JVM mode only.
```
rsvoboda, line 13:
```suggestion
  * This affects JVM mode only.
```

QUARKUS-1936.md
```diff
@@ -0,0 +1,31 @@
...
+This document is a high level test plan for certification of Red Hat build of Quarkus on Temurin JVM.
+
+## Scope of the testing
+
+### Impact on test suites and testing automation
+* Temurin JVM 11 and 17 need to be added as JDKs to Quarkus QE Jenkins
+  * Temurin JVM 11 and 17 seem to be updated timely on `/qa/tools` mount point, so that will be used for the Elektra
+    time frame.
+* Temurin JVM 11 and 17 have to be added as JVMs to Polarion
+* Bare metal and OpenShift jobs in extended platform testing need to be extended to be ran with both Temurin 11 and 17.
```
rsvoboda, line 12: How will be OpenShift testing affected? Do we assume new base image instead of OpenJDK one? 
Or just app compiled using Temurin but running in OpenJDK based pod?

TBH I don't see much value in regular OpenShift testing with Temurin if there is no Temurin base image we support

QUARKUS-1936.md
```diff
@@ -0,0 +1,31 @@
...
+* This affects JVM mode only.
+
+### Impact on resources
+* 1 medium executor for bare metal testing for 4 hours on Temurin 11
```
rsvoboda, line 16: we do not use "medium" flavor, I think we end up using large in the test-suite jobs

QUARKUS-1936.md
rsvoboda: https://issues.redhat.com/browse/QUARKUS-1936

> Note that there are no UBI Java images based on Adoptium so we are only targeting bare metal at this point.

rsvoboda (changes requested):

rsvoboda (approved):

PR #84, opened 2022-07-05
Add QUARKUS-1325 - Helm chart support for Quarkus

pjgg:
Quarkus Helm chart test-plan

- Generate and deploy a Quarkus Helm Chart over Openshift default repository
- Update a deployed chart with a new Quarkus application chart version
- Verify some mapping between Quarkus configuration and generated Helm chart:
  - Quarkus SmallRye Health

### DIFF
--- a/QUARKUS-1325.md
+++ b/QUARKUS-1325.md
@@ -1,0 +1,35 @@
+# QUARKUS-1325 - Helm chart support for Quarkus
+
+JIRA link: https://issues.redhat.com/browse/QUARKUS-1325
+
+Quarkus Guide: https://quarkiverse.github.io/quarkiverse-docs/quarkus-helm/dev/index.html
+
+This test plan cover Quarkus - Helm integration
+
+## Scope of the testing
+
+Verify a repeatable/versioned way to deploy Quarkus application into K8s/ocp through helm. We are assuming that the user is going to follow the Helm standard release flow, which means that the user will push his Quarkus app to his configured Docker registry and then this docker image is going to be used by the generated `chart.yaml`. This `build app -> push app to docker registry -> Helm chart generation` is done through `quarkus-helm` + `quarkus-container-image-docker` as is explained in Quarkus Helm extension [documentation](https://quarkiverse.github.io/quarkiverse-docs/quarkus-helm/dev/index.html). 
+
+The deployment to K8s/OCP is done as a regular Helm chart. Ex `helm -f app-chart.yaml sync`
+
+Test development will focus on
+- JVM mode. We have internal conversations about the convenience of verifying JVM and native executions and we've concluded that Helm is concerned just about image name and doesn't care if the app is JVM or Native based. For this reason and in order to save resources we are only going to cover JVM mode.
+- Generate and deploy a Quarkus Helm Chart over OpenShift default repository
+- Deploy app container image into quay.io
+- Verify some mapping between Quarkus configuration and generated Helm chart:
+  - Quarkus SmallRye Health 
+
+### Impact on testsuites and testing automation:
+
+  - A new module will be created on Quarkus test-suite in order to cover this scenarios
+  - Jenkins slaves instances will have Helm / helmfile installed and will deploy a Quarkus application through Helm into an OpenShift cluster instance. This is a new way to deploy into K8s/OpenShift that will increase slightly the total execution of time of Quarkus test-suite. Native execution should be avoided, because we don't see any additional value for these scenarios.  
+  - Charts should be removed from OpenShift default repository after test execution 
+  
+## Getting familiar with the feature
+Following actions were taken to ensure familiarity:
+ - Exploratory testing with current Quarkiverse extension
+ - Ensure upstream CI coverage
+ - Ensure good user experience and simplicity of use
+
+## Contacts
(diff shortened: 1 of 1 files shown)


rsvoboda:
https://issues.redhat.com/browse/QUARKUS-1325 has poor details about the requirement, it has just 2 lines.

Push on providing more details in the description of the requirement, without these details we can't properly review the test plan.

I miss details about the intended use-cases and intended flow, level of support - full support / TP (it's quirkiverse extension, not even in the platform BOM), both JVM and Native mode are adequately supported / no limitations, additional tooling requirements, documentation requirements on community and product side ...

rsvoboda:
Added Josef instead of Michal as Josef acts as backup for Helm topic

QUARKUS-1325.md
```diff
@@ -0,0 +1,27 @@
...
+JIRA link: https://issues.redhat.com/browse/QUARKUS-1325
+
+This test plan cover Quarkus - Helm integration
+
+## Scope of the testing
+
+Verify a repeatable/versioned way to deploy Quarkus application into K8s/ocp through helm
+
+Test development will focus on
+- Generate and deploy a Quarkus Helm Chart over Openshift default repository
```
rsvoboda, line 12: Will you also cover deploying to external registry like quay.io?

QUARKUS-1325.md
```diff
@@ -0,0 +1,27 @@
...
+### Impact on testsuites and testing automation:
+  - Github CI jobs will be excluded from Helm verification. 
+  - Jenkins slaves instances will have Helm / helmfile installed and will deploy a Quarkus application through Helm into an Openshift cluster instance. This is a new way to deploy into K8s/Openshift that will increase slightly the total execution of time of Quarkus Test-suite.  
+  - Charts should be removed from Openshift default repository after test execution 
+  
+## Getting familiar with the feature
+Following actions were taken to ensure familiarity:
+ - Exploratory testing with current Quarkiverse extension
+ - Ensure good user experience and simplicity of use
+
```
rsvoboda, line 25: I would add something like "Ensure upstream CI coverage" to check if the extension has CI against Kubernetes.

QUARKUS-1325.md
```diff
@@ -0,0 +1,27 @@
+# QUARKUS-1325 - Helm chart support for Quarkus
+
+JIRA link: https://issues.redhat.com/browse/QUARKUS-1325
+
+This test plan cover Quarkus - Helm integration
+
+## Scope of the testing
```
rsvoboda, line 7: Openshift => OpenShift  (4 occurrences)
Quarkus Test-suite => Quarkus test-suite (1 occurrence)

QUARKUS-1325.md
```diff
@@ -0,0 +1,27 @@
...
+
+Verify a repeatable/versioned way to deploy Quarkus application into K8s/ocp through helm
+
+Test development will focus on
+- Generate and deploy a Quarkus Helm Chart over Openshift default repository
+- Verify some mapping between Quarkus configuration and generated Helm chart:
+  - Quarkus SmallRye Health 
+
```
rsvoboda, line 15: Will you ensure both JVM and Native applications get deployed?

rsvoboda (changes requested):

QUARKUS-1325.md
pjgg: No, for this first release I was thinking to cover the default scenario which means: 
- Deploy app container image into docker.io | quay.io 
- install Helm charts into the (OpenShift) default helm chart repository

QUARKUS-1325.md
pjgg: That is a good question... I think that JVM would be enough because, from Helm side point of view, they don't care about native or JVM, Helm will install a Chart that will point to a docker container registry that has the app image (Native or JVM)

So... I think that we should only cover JVM mode because from my point of view native mode is more expensive in terms of resources and doesn't add any value to this scenario.

WDYT?

QUARKUS-1325.md
rsvoboda: Pls add `Deploy app container image into quay.io` detail

QUARKUS-1325.md
rsvoboda: From testing perspective it seems to be fine to focus just on JVM mode.

We should probably consider what's the flow for users, do they provide just image in one mode or they do both `abc-jvm` and `abc-native` images. This is not exclusively tight to helm topic, it's more generic topic for pushing the images from Quarkus based project. My view is that usually one mode gets picked and that' it. Dual image push may not be the common case.

But pls comment about this in TP to indicate this topic was considered + add the reasoning.
`For this first coverage we don't think that native mode is relevant.` is too generic, please indicate that Helm is concerned just about image name and doesn't care if the app is JVM or Native based.

QUARKUS-1325.md
rsvoboda: This is still not fixed!

QUARKUS-1325.md
rsvoboda: This is still not fixed!

rsvoboda (changes requested):

rsvoboda (approved):

rsvoboda:
Josef is sick, one review should be enough, merging.

PR #93, opened 2022-09-13
QUARKUS-2072 - Defined boundaries for supported dependencies in RHBQ

rsvoboda:
QUARKUS-2072 - Defined boundaries for supported dependencies in RHBQ


mjurc (approved):
LGTM, thank you for the update.

PR #107, opened 2022-10-26
QUARKUS-2489: Add logout flow on OIDC

pjgg:
Backchannel logout applies to `quarkus-oidc` and basically is a standard way to do logout in to a single sign-on service.
We will use Keycloak as an OIDC provider and the flavor of the extensions (classic or reactive) is not relevant for this coverage.

### DIFF
--- a/QUARKUS-2489.md
+++ b/QUARKUS-2489.md
@@ -1,0 +1,36 @@
+# QUARKUS-2489 OIDC Back channel logout
+
+Jira: https://issues.redhat.com/browse/QUARKUS-2489
+
+This feature provides Back-Channel Logout capability for OIDC extension.
+
+## Scope of the testing
+
+Backchannel logout applies to `quarkus-oidc` and basically is a standard way to logout the current user from all the applications this user is currently logged in, bypassing the user agent.
+
+We will create a valid token and proceed with the logout flow.
+### Impact on test suites and testing automation
+
+We will need to cover JVM and native mode in order to double-check that there are no conflicts with the native compilations. New Keycloak configuration id need it with all the logout flow parameters. 
+### Impact on resources:
+
+The new scenario is going to run in JVM and native mode over baremetal and OpenShift. So we are expecting about 5 min per module on JVM mode, and 10 min per module on native mode. Running a test over OpenShift used to take some extra time but is just an aproximation. To sum up, we are expecting an increase of 30 min of total amount execution.
+## Getting familiar with the feature
+
+Following actions were taken to ensure familiarity:
+- Ensure documentation provides a clear explanation of expected behavior
+- Review upstream and quickstarts coverage
+- Read OpenID Connect specifications
+- Manual testing
+
+## Contacts
+
+* Tester: Pablo Gonzalez Granados <pagonzal@redhat.com>
+
+## References
+
+- [Quarkus OpenID Connect documentation](https://quarkus.io/guides/security-openid-connect-web-authentication#back-channel-logout)
+- [OpenID Connect 1.0 spec](https://openid.net/specs/openid-connect-backchannel-1_0.html#Backchannel)
+- [OIDC logout PR](https://github.com/quarkusio/quarkus/pull/24611)
+- [Keycloak logout](https://www.keycloak.org/docs/latest/server_admin/index.html#_oidc-logout)
+  


QUARKUS-2489.md
```diff
@@ -0,0 +1,53 @@
+# QUARKUS-2489 OIDC Back channel logout
+
+Jira: https://issues.redhat.com/browse/QUARKUS-2489
+
+This feature is talking about logout flow on an OIDC provider as Keycloak
+
+## Scope of the testing
+
+Backchannel logout applies to `quarkus-oidc` and basically is a standard way to do logout in to a single sign-on service.
+We will use Keycloak as an OIDC provider and the flavor of the extensions (classic or reactive) is not relevant for this coverage.
```
rsvoboda, line 10: you mean resteasy extensions?

QUARKUS-2489.md
```diff
@@ -0,0 +1,53 @@
+# QUARKUS-2489 OIDC Back channel logout
+
+Jira: https://issues.redhat.com/browse/QUARKUS-2489
+
+This feature is talking about logout flow on an OIDC provider as Keycloak
+
+## Scope of the testing
+
+Backchannel logout applies to `quarkus-oidc` and basically is a standard way to do logout in to a single sign-on service.
```
rsvoboda, line 9: logout from a single sign-on service?

QUARKUS-2489.md
```diff
@@ -0,0 +1,53 @@
+# QUARKUS-2489 OIDC Back channel logout
+
+Jira: https://issues.redhat.com/browse/QUARKUS-2489
+
+This feature is talking about logout flow on an OIDC provider as Keycloak
```
rsvoboda, line 5:
```suggestion
This feature provides Back-Channel Logout capability for OIDC extension.
```

QUARKUS-2489.md
```diff
@@ -0,0 +1,53 @@
...
+- Validate the iss, aud, and iat Claims in the same way they are validated in ID Tokens
+- Verify that the Logout Token contains a sub Claim, a sid Claim, or both
+- Verify that the Logout Token contains an events Claim whose value is JSON object containing the member name http://schemas.openid.net/event/backchannel-logout
+- Verify that the Logout Token does not contain a nonce Claim.
+  
+If any of the validate logout tokens fails you must return an HTTP 400. Otherwise, proceed to perform the logout actions.
+
+### Impact on test suites and testing automation
+
+A New logout endpoint is going to be created in some of the existing OIDC scenario, for example `keycloak-oidc-client-basic`, and we will need to cover JVM and native mode in order to double-check that there are no conflicts with the native compilations.
```
rsvoboda, line 29: Do not extend the basic module, it definitely fits into advanced on

And please cover both keycloak-oidc-client-extended and keycloak-oidc-client-reactive-extended so the module are in sync as much as possible

QUARKUS-2489.md
```diff
@@ -0,0 +1,53 @@
...
+- Verify that the Logout Token contains an events Claim whose value is JSON object containing the member name http://schemas.openid.net/event/backchannel-logout
+- Verify that the Logout Token does not contain a nonce Claim.
+  
+If any of the validate logout tokens fails you must return an HTTP 400. Otherwise, proceed to perform the logout actions.
+
+### Impact on test suites and testing automation
+
+A New logout endpoint is going to be created in some of the existing OIDC scenario, for example `keycloak-oidc-client-basic`, and we will need to cover JVM and native mode in order to double-check that there are no conflicts with the native compilations. 
+
+No OpenShift coverage is required for this scenario.
```
rsvoboda, line 31: Pls explain in the text why OpenShift scenario is not needed.

OpenShift scenarios run with RH-SSO, so I would say it's worth checking

QUARKUS-2489.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Impact on test suites and testing automation
+
+A New logout endpoint is going to be created in some of the existing OIDC scenario, for example `keycloak-oidc-client-basic`, and we will need to cover JVM and native mode in order to double-check that there are no conflicts with the native compilations. 
+
+No OpenShift coverage is required for this scenario. 
+
+### Impact on resources:
+
+This scenario will require deploying a Keycloak and running some extra tests over an existing module. So, I would say that the total amount of time is going to be around 3 min extra minutes in JVM mode and 5 min in native mode.
```
rsvoboda, line 35: Pls update estimates for changed scope (2 modules, openShift)

QUARKUS-2489.md
```diff
@@ -0,0 +1,53 @@
...
+This feature is talking about logout flow on an OIDC provider as Keycloak
+
+## Scope of the testing
+
+Backchannel logout applies to `quarkus-oidc` and basically is a standard way to do logout in to a single sign-on service.
+We will use Keycloak as an OIDC provider and the flavor of the extensions (classic or reactive) is not relevant for this coverage. 
+
+Absolute Back-Channel Logout URL is calculated by adding `quarkus.oidc.back-channel-logout.path` to the current endpoint URL and then based on your configuration you could do a local and / or a global logout.
+
+We will create a valid token by an authorization flow and then after validating the token against some authenticated API we will proceed with the logout flow and verify the following constraints:
```
rsvoboda, line 14: Is the following test coverage really necessary?

Aren't you testing Keycloak functionality instead of testing of the integration?

QUARKUS-2489.md
```diff
@@ -0,0 +1,53 @@
...
+
+This scenario will require deploying a Keycloak and running some extra tests over an existing module. So, I would say that the total amount of time is going to be around 3 min extra minutes in JVM mode and 5 min in native mode.
+
+## Getting familiar with the feature
+
+Following actions were taken to ensure familiarity:
+- Ensure documentation provides a clear explanation of expected behavior
+- Review upstream and quickstarts coverage
+- Read openID connect specifications
+
```
rsvoboda, line 43: Add point for manual experiments to see how the feature works from end user perspective, not just in tests.

rsvoboda (changes requested):

rsvoboda:
Requesting additional review from @michalvavrik as he worked on several security / oidc topics.

QUARKUS-2489.md
michalvavrik: While OIDC layer that handles this before RESTEasy classic/reactive, what happens when things goes wrong is very much not. F.e. PR added code that throws `InvalidJwtException` (it's caught later, but that's implementation detail that can easily change), and the extensions handle these situations via their own mechanisms. IMHO complex test for Reactive is fine, but POC for classic (some minimal scenario) would be nice.

QUARKUS-2489.md
michalvavrik: I think this comes from specs chapter `2.6.  Logout Token Validation` right :-) So there are 2 situations we need to test:

- Back channel logout goes to Quarkus app and the app sends it to the Keycloak, in such case the response is always going to be empty (see `io.quarkus.oidc.runtime.BackChannelLogoutHandler.RouteHandler`), so just validate response code.
- User logged out in different app (it can be Quarkus app too), and Keycloak informed the tested Quarkus app that logging has happened. Here is where actual value of our tests lay as upstream is not testing this AFAICT. If we don't test it, we can't tell if we are logged out based on Keycloak call back are Quarkus performed local logout. In this case, there is no logout token.

So IMHO no place for logout token validation.

QUARKUS-2489.md
michalvavrik: +1

QUARKUS-2489.md
michalvavrik: I don't agree. Huge portion of SPAs are going to logout directly (call SSO). TBH all Angular apps I saw do that, but I'm sure there are people who don't do that. Therefore what we need to do is test that we actually get callback from Keycloak when logout happen and Quarkus log us out.

QUARKUS-2489.md
```diff
@@ -0,0 +1,53 @@
...
+- Refresh tokens issued without the offline_access property to a session being logged out SHOULD be revoked. Refresh tokens issued with the offline_access property normally SHOULD NOT be revoked.
+- Logout token should be encrypted, decrypt it using the keys and algorithms that the Client specified during Registration. If logout Token is not encrypted, then reject it
+- Validate the Logout Token signature in the same way that an ID Token signature is validated
+- Validate the algorithm Header Parameter in the same way it is validated for ID Tokens
+- Validate the iss, aud, and iat Claims in the same way they are validated in ID Tokens
+- Verify that the Logout Token contains a sub Claim, a sid Claim, or both
+- Verify that the Logout Token contains an events Claim whose value is JSON object containing the member name http://schemas.openid.net/event/backchannel-logout
+- Verify that the Logout Token does not contain a nonce Claim.
+  
+If any of the validate logout tokens fails you must return an HTTP 400. Otherwise, proceed to perform the logout actions.
```
michalvavrik, line 25: Logout is async, only time you get 400 is that your token claims are invalid (e.g. missing back-channel claim) or expiration is missing or token is expired. We could check that as it makes sense to require valid token, but IMHO not very important as it's something like "pre-validaton", actual validation is done by Keycloak and we won't know about that except that we were not logged out. You decide :-)

michalvavrik (changes requested):

QUARKUS-2489.md
michalvavrik: Sentence is not wrong though, maybe I concentrated on word "standard" to much.

QUARKUS-2489.md
michalvavrik: Except to what you suggest below - that if we send Quarkus app invalid logout token, we get 400.

michalvavrik: Although both Keycloak and us calls same URL that keeps back-channel log out token, so there should be a little difference to who calls. IMHO we should test full flow anyway.

QUARKUS-2489.md
rsvoboda: What I mean is this, shouldn't be the sentence changed this way?

"standard way to do logout in to a single sign-on service"
 =>
"standard way to do logout out from a single sign-on service"

QUARKUS-2489.md
rsvoboda: thanks for input, this makes sense, there should be some coverage for resteasy classic

QUARKUS-2489.md
rsvoboda: just to be clear @michalvavrik , you vote for full coverage described here?

michalvavrik: no @rsvoboda I think it doesn't make sense. We should call back-channel logout endpoint at least once with invalid token so we do that Sergey implemented checks correctly (according to OIDC specs), but I'd not validate logout token myself as that's not our job.

michalvavrik: by ^^^ I tried to point out that I'm missing in TP bound that we will test callback from KC.

QUARKUS-2489.md
pjgg: Yes, the main goal of the logout endpoint is to remove all the active sessions in all devices for a given user (except that the user is using an offline token). 
Ref: https://github.com/quarkusio/quarkus/pull/24611/files#diff-75b29fe10300c4376f22f9584de09fe4d6d993f31fb218d3966ccb9d30c8608bR474

pjgg: > maybe I concentrated on wo

Standard ref that Quarkus must implement: https://openid.net/specs/openid-connect-backchannel-1_0.html#Backchannel

michalvavrik: @pjgg you are right that these constraints need to be implemented by Quarkus (RP), but in order to be sure that Quarkus enforce all of these constraints you would have to call Quarkus for each of these scenarios (e.g. add nonce claim) and check response code as that's your only indicator whether Quarkus validated it. You can't do it in  bulk. These seems a lot like unit tests rather than integration tests. Do we test in such details all other features? If I'm wrong, please just explain it to me. AFAIK I didn't see it in TS.

QUARKUS-2489.md
```diff
@@ -0,0 +1,36 @@
...
+A New logout endpoint is going to be created in `keycloak-oidc-client-extended` and `keycloak-oidc-client-reactive-extended`, and we will need to cover JVM and native mode in order to double-check that there are no conflicts with the native compilations. 
+### Impact on resources:
+
+The new scenario is going to run in JVM and native mode over baremetal and OpenShift. So we are expecting about 5 min per module on JVM mode, and 10 min per module on native mode. Running a test over OpenShift used to take some extra time but is just an aproximation. To sum up, we are expecting an increase of 30 min of total amount execution.
+## Getting familiar with the feature
+
+Following actions were taken to ensure familiarity:
+- Ensure documentation provides a clear explanation of expected behavior
+- Review upstream and quickstarts coverage
+- Read openID connect specifications
```
michalvavrik, line 23: Please rewrite to _openID connect_ to _OpenID Connect_.

QUARKUS-2489.md
```diff
@@ -0,0 +1,36 @@
...
+This feature provides Back-Channel Logout capability for OIDC extension.
+
+## Scope of the testing
+
+Backchannel logout applies to `quarkus-oidc` and basically is a standard way to logout the current user from all the applications this user is currently logged in, bypassing the user agent.
+
+We will create a valid token by an authorization flow and proceed with the logout flow.
+### Impact on test suites and testing automation
+
+A New logout endpoint is going to be created in `keycloak-oidc-client-extended` and `keycloak-oidc-client-reactive-extended`, and we will need to cover JVM and native mode in order to double-check that there are no conflicts with the native compilations.
```
michalvavrik, line 14: > A New logout endpoint is going to be created in `keycloak-oidc-client-extended` and `keycloak-oidc-client-reactive-extended`

Can you please explain (in comment is enough) why would you create new logout endpoint? AFAICT Quarkus has handler that expects logout token on the logout back-channel path , so there are 2 possible situations:

- you call RP that is also OP (KC?), and KC calls the tested app (RP)
- back-channel logout in the tested app (Quarkus provides route handler itself)

In neither of these scenarios, you need the logout endpoint. Maybe you mean some endpoint that checks you are logged out (if we don't have such endpoint? please check)?

QUARKUS-2489.md
```diff
@@ -0,0 +1,36 @@
...
+
+Jira: https://issues.redhat.com/browse/QUARKUS-2489
+
+This feature provides Back-Channel Logout capability for OIDC extension.
+
+## Scope of the testing
+
+Backchannel logout applies to `quarkus-oidc` and basically is a standard way to logout the current user from all the applications this user is currently logged in, bypassing the user agent.
+
+We will create a valid token by an authorization flow and proceed with the logout flow.
```
michalvavrik, line 11: Who you gonna call? Quarkus app or KC or ..?

michalvavrik (changes requested):
I left my opinions, please also double-check with @rsvoboda . Thank you for the changes.

QUARKUS-2489.md
pjgg: You could add a `post-logout` endpoint, let's make it the most simple as we can, and maybe in the future, we can make a more advance scenario. So +1 to remove it!.

QUARKUS-2489.md
pjgg: We are gonna make a request to an authenticated endpoint that will forward the request to KC, once the login is done we will verify that the session was created and then proceed with the logout. And re-check again that the session was removed.

michalvavrik: Thanks for explanation.

QUARKUS-2489.md
michalvavrik: @pjgg Changes look good, but what about openID connect, I requested changes as all linked references use OpenID Connect. If you find request for change unreasonable, please explain why. Thank you

michalvavrik (approved):
lgtm

PR #115, opened 2022-11-30
QUARKUS-2246 - Integrate Vertx HTTP extension with custom CredentialsProvider

rsvoboda:
QUARKUS-2246 - Integrate Vertx HTTP extension with custom CredentialsProvider

https://issues.redhat.com/browse/QUARKUS-2246

### DIFF
--- a/QUARKUS-2246.md
+++ b/QUARKUS-2246.md
@@ -1,0 +1,19 @@
+# QUARKUS-2246 - Integrate Vertx HTTP extension with custom CredentialsProvider
+
+JIRA: https://issues.redhat.com/browse/QUARKUS-2246
+
+Integrate Vertx HTTP extension with custom CredentialsProvider, main motivation is that RHPAM product team can use FileVaultCredentialsProvider to support their use cases.
+
+RHBQ product team provides the means to use custom CredentialsProvider.
+
+## Scope of the testing
+
+- Review the PR introducing the functionality
+- Evaluate the test coverage available in upstream
+- Check if additional test coverage is needed
+- Implement the additional test coverage
+
+The primary target should be `integration-tests` module of the upstream Quarkus repo.
+
+### Impact on resources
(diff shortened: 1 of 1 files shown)


michalvavrik (approved):
I reviewed upstream PRs and docs and I am okay with the test plan.

Please consider following topics, but you don't have to add them to the TP:

- we need to add an example into docs or open an issue for that as if I didn't see the implementation, I wouldn't be able to explain how `io.quarkus.vertx.http.runtime.CertificateConfig#keyStoreKeyPassword` is used to retrieve passwrod from `CredentialsProvider`
- it would be nice to have tests that has conflicts that guarantees behavior explained in docs. IMO you won't find bug, but it's weird there is no unit test coverage for this as such tests makes contract
  -  key store password set in both provider and properties file
  - both password and password key are set, make sure password has higher priority
  - named credentialsprovider
  - I think current implementation does not validate that named CredentialsProvider set in application.properties exists, that seems lame to me as you can easily do it and thus provide extra information to user at build time (and prevent runtime errors)
  - You should also fix related typos, e.g.
    - extra curly brace https://github.com/sberyozkin/quarkus/blob/main/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/CertificateConfig.java#L20
    - missing dot https://github.com/quarkusio/quarkus/blame/main/docs/src/main/asciidoc/http-reference.adoc#L137

rsvoboda:
It's intentional to not promote CredentialsProvider path, MP / SmallRye Config is the preferred way

rsvoboda:
> Also it has to be `@Unremovable` which would be nice to have mentioned in docs.

There is example in https://quarkus.io/guides/credentials-provider#new-credentials-provider-extension with `@Unremovable`

michalvavrik:
I take it back, thanks.

PR #120, opened 2022-12-23
Add test plan for QUARKUS-2744

michalvavrik:
Adds a test plan for https://issues.redhat.com/browse/QUARKUS-2744.

### DIFF
--- a/QUARKUS-2744.md
+++ b/QUARKUS-2744.md
@@ -1,0 +1,48 @@
+# QUARKUS-2744 - Handling a multipart request part as a file based on the content-type
+
+JIRA: https://issues.redhat.com/browse/QUARKUS-2744
+
+Quarkus documentation: https://quarkus.io/guides/resteasy-reactive#multipart
+
+PR: https://github.com/quarkusio/quarkus/pull/29729
+
+Upstream issue: https://github.com/quarkusio/quarkus/issues/29725
+
+Ability to indicate whether a given multipart field should be handled as a file part based on configured content type has been added in Quarkus 2.16 and backported to Quarkus 2.13.6.
+This change allows to access the parts as files even though file form parameter name is not known beforehand.
+
+## Scope of testing
+
+Tests should verify:
+
+- file upload is accessible if the file part has a standard content type set with `quarkus.http.body.multipart.file-content-types` configuration property
+- file upload is accessible if the file part has a custom content type set with `quarkus.http.body.multipart.file-content-types` configuration property
+- any field of form data container class annotated with `@RestForm` with data type `java.io.File` has access to same-named file part
+
+### Existing tests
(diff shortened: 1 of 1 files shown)


QUARKUS-2744.md
```diff
@@ -0,0 +1,45 @@
...
+Upstream issue: https://github.com/quarkusio/quarkus/issues/29725
+
+Ability to indicate whether a given multipart field should be handled as a file part based on configured content type has been added in Quarkus 2.16 and backported to Quarkus 2.13.6.
+This change allows to access the parts as files even though file form parameter name is not known beforehand.
+
+## Scope of testing
+
+Tests should verify:
+
+- file upload is accessible if the file part has standard content type set with `quarkus.http.body.multipart.file-content-types` configuration property
```
rsvoboda, line 15:
```suggestion
- file upload is accessible if the file part has a standard content type set with `quarkus.http.body.multipart.file-content-types` configuration property
```

QUARKUS-2744.md
```diff
@@ -0,0 +1,45 @@
...
+- file upload is accessible if the file part has custom content type set with `quarkus.http.body.multipart.file-content-types` configuration property
```
rsvoboda, line 16:
```suggestion
- file upload is accessible if the file part has a custom content type set with `quarkus.http.body.multipart.file-content-types` configuration property
```

rsvoboda (changes requested):
Small typos to be fixed

One question - Is chunked transfer encoding something to consider here? E.g. because content-length is unknown / not set.

michalvavrik:
> Small typos to be fixed 

Thank you.

> One question - Is chunked transfer encoding something to consider here? E.g. because content-length is unknown / not set.

If I thought about it sooner I'd probably add it just in case I'm wrong, but there are technical reasons why it is IMO not relevant:

- I agree with thinking that amounts of data may not be known until the request has been fully processed with this feature, but technically situation is identical to file names as you don't know how large amount of data is coming there either
- chunked must be used when message length cannot be determined, but that's no change either, we simply send form data and server don't know how we are going to name the fields
- (not sure here!) I don't think Quarkus can enforce certain request Transfer-Encoding, isn't it by nature client option? (set headers or not). The header should imply what type of transformation has been applied to the message body
- I think transfer encoding is handled by Vert.x because now that I was looking into RR code, I can't find anything relevant
- chunked transfer encoding is how you transfer data encode whole message, it happens on data transfer level and for whole message, therefore it must work for both parts with/without known file names and independently on content types
- we could summarize changes than like: before we had one IF with _file name matches_ and now there is OR with _it's one of media types in config_, changes could not affect how it worked till now

Based on your question I'd probably have another one: how can I limit total content-length of files? It seems it is possible if you write request filter that checks headers.

rsvoboda:
> How can I limit total content-length of files? 

https://quarkus.io/guides/all-config#quarkus-vertx-http_quarkus.http.limits.max-body-size should be sufficient?

rsvoboda (approved):

PR #126, opened 2023-01-18
Add QUARKUS-2746: ConsumeEvent annotation: Duplicated event consumption

pjgg:
ConsumeEvent annotation should be invoked once when generic parent class
has an abstract method for it.

Documentation: https://quarkus.io/guides/reactive-event-bus#consuming-events


pjgg:
Bugfixes don't require a TP.

PR #130, opened 2023-01-25
Tests plans for Micrometer metrics inside managed Vert.X

fedinskiy:
### Links

JIRA: https://issues.redhat.com/browse/QUARKUS-2829

Quarkus documentation: https://quarkus.io/guides/vertx

### Checklist of areas related to the topic

Select all applicable areas that have been considered in this test plan:

- [ ] Logging
- [ ] Tracing
- [x] Metrics
- [ ] Security
- [ ] OpenAPI
- [x] Data sources
- [ ] Qute

### DIFF
--- a/QUARKUS-2829.md
+++ b/QUARKUS-2829.md
@@ -1,0 +1,56 @@
+# QUARKUS-2829 - Internal support for Micrometer metrics when using managed Vert.x in Quarkus
+
+JIRA: https://issues.redhat.com/browse/QUARKUS-2829
+
+
+As part of supporting Vert.x use-cases on Quarkus we need to support collecting metrics using micrometer for the managed Vert.x instance in Quarkus. This is also a requirement for some layered products like AMQ Streams and OpenShift Serverless that today are using Vert.x, but needs to move to use Quarkus. For 2.13 the support for this is limited to OpenShift Serverless and AMQ Streams, this is not supported for general use. Support for general use should come in RHBQ 3.x
+
+## Scope  of the implementation
+Managed Vert.x in Quarkus contains following extensions:
+- Messaging clients:
+  AMQP,Kafka, MQTT. Extension names: quarkus-smallrye-reactive-messaging-amqp, *-kafka, *-mqtt.
+- DB clients: Mssql, DB2, Oracle, MySQL, Postgres(pg). Extension names: quarkus-reactive-(name)-client.
+- Redis client: quarkus-redis-client
+- quarkus-mailer
+- quarkus-vertx (the core API)
+
+Current implementation provides following metrics (full names is `poolType+".pool."+MetricName`):
+- queue.delay
+- queue.size
+- usage
+- active
+- idle
+- ratio
+- completed
+- processing
+- current
+- reset
+- pending
+- connections
+
+## Scope of testing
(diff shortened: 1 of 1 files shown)


QUARKUS-2829.md
```diff
@@ -0,0 +1,55 @@
...
+
+JIRA: https://issues.redhat.com/browse/QUARKUS-2829
+
+
+As part of supporting Vert.x use-cases on Quarkus we need to support collecting metrics using micrometer for the managed Vert.x instance in Quarkus. This is also a requirement for some layered products like AMQ Streams and OpenShift Serverless that today are using Vert.x, but needs to move to use Quarkus. For 2.13 the support for this is limited to OpenShift Serverless and AMQ Streams, this is not supported for general use. Support for general use should come in RHBQ 3.x
+
+## Scope of testing
+
+Managed Vert.x in Quarkus contains following extensions:
+- Messaging clients:
```
mjurc, line 11: The messaging clients are not supported, at least not in RHBQ 2.13.

If we're listing those extensions, maybe it would be useful bit of information they're not supported.

QUARKUS-2829.md
```diff
@@ -0,0 +1,55 @@
...
+- ratio
+- completed
+- processing
+- current
+- reset
+- pending
+- connections
+
+Ideally (for Quarkus 3.0), we need to check, that we can get metrics for all supported extensions, and for working with Vert.X specific functionality of verticles and event bus.
+According to Thomas, the main focus of the current change is Serverless. Therefore, the optimal way is to cover only basic functionality for applications deployed in serverless mode.
```
mjurc, line 34: According to PM may be better for people who don't know who Thomas is.

I also think that this information would be better at the start of "Scope of testing" section, as this is the actual intended scope. What do you think?

To sum it up in the way I understand it - the scope of QUARKUS-2829 is limited internal support for current internal Vert.x users - namely Serverless and AMQ streams. We are going to test just the serverless scenarios.

For full support, we will need to implement the coverage for the rest of the extensions you provided, but that's not the scope of this issue - so maybe that would be more suitable in future consideration.

QUARKUS-2829.md
```diff
@@ -0,0 +1,55 @@
...
+- idle
+- ratio
+- completed
+- processing
+- current
+- reset
+- pending
+- connections
+
+Ideally (for Quarkus 3.0), we need to check, that we can get metrics for all supported extensions, and for working with Vert.X specific functionality of verticles and event bus.
```
mjurc, line 33: Maybe this should be in future considerations chapter - this won't be in scope of testing for QUARKUS-2829, since that's scoped just for what you mention in the next sentence.

mjurc (changes requested):

QUARKUS-2829.md
fedinskiy: Moved to "future considerations"

QUARKUS-2829.md
mjurc: `quarkus-smallrye-reactive-messaging-kafka` is supported, so is `quarkus-smallrye-reactive-messaging-amqp`. MQTT is not supported:

https://code.quarkus.redhat.com/?extension-search=origin:platform%20quarkus-smallrye-reactive-messaging-mqtt

mjurc (approved):
Thanks for the updates!

PR #146, opened 2023-08-01
QUARKUS-3290 - Generate source zip which contains the source code of …

rsvoboda:
QUARKUS-3290 - Generate source zip which contains the source code of the built artifacts

JIRA: https://issues.redhat.com/browse/QUARKUS-3290

Quarkus documentation: not applicable

   - Logging
   - Tracing
   - Metrics
   - Security
   - OpenAPI
   - Data sources
   - Frontend
   - Qute

### DIFF
--- a/QUARKUS-3290.md
+++ b/QUARKUS-3290.md
@@ -1,0 +1,38 @@
+# QUARKUS-3290 - Generate source zip which contains the source code of the built artifacts
+
+JIRA: https://issues.redhat.com/browse/QUARKUS-3290
+
+As part of discussions abound QUARKUS-2701, it was decided to generate source zip which contains the source code of the built Maven artifacts, sources will be pulled from the internal Gerrit repository for upcoming 3.x releases.
+
+One of the reasons is to fully comply with Complete Corresponding Source Code recommendation from Red Hat compliance training [1].
+
+Another big reason is to demonstrate the amount of work that is done during the productization effort - should help PM in his discussions and with the secure build chain efforts.
+
+## Scope of the testing
+ * Exploratory testing
+ * Evaluation of existing test scenarios
+ * New test development
+
+Topics for exploratory testing:
+ * SRC ZIP naming
+ * Root directories naming
+ * Sources for platform participants
+ * Review of included sources for transitive dependencies
+ * Files count, summary per file extension
+ * Large files check
+ * Empty files and directories check
+
+### Impact on test suites and testing automation
+ * Marete checks for source zip need to be reworked
+   * Adjustments of the tests based on the experience from the exploratory testing and feedback in filed JIRAs
+     * These changes need to be communicated with other Marete users
+   * Adjustments of the configuration based on the experience from the exploratory testing and feedback in filed JIRAs
+
+### Impact on resources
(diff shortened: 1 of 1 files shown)


QUARKUS-3290.md
```diff
@@ -0,0 +1,35 @@
...
+ * Evaluation of existing test scenarios
+ * New test development
+
+Topics for exploratory testing:
+ * SRC ZIP naming
+ * Root directories naming
+ * Sources for platform participants
+ * Review of included sources for transitive dependencies
+ * Files count, summary per file extension
+ * Large files check
```
fedinskiy, line 22: Just curious: why check large files?

QUARKUS-3290.md
```diff
@@ -0,0 +1,35 @@
+# QUARKUS-3290 - Generate source zip which contains the source code of the built artifacts
+
+JIRA: https://issues.redhat.com/browse/QUARKUS-3290
+
+As part of discussions abound QUARKUS-2701, it was decided to generate source zip which contains the source code of the artifact from the internal Gerrit repository for upcoming 3.x release
```
fedinskiy, line 5: "The artifact" means our main zip for binaries?

QUARKUS-3290.md
```diff
@@ -0,0 +1,35 @@
...
+
+One of the reasons is to fully comply with https://issues.redhat.com/secure/attachment/12796750/screenshot-1.png from Red Hat compliance training
```
fedinskiy, line 7: Do we have this text in writing somewhere? Link to jira attachment with text isn't the best solution

fedinskiy (commented):
Just a couple of questions

QUARKUS-3290.md
rsvoboda: e.g. https://issues.redhat.com/browse/QUARKUS-3293 ;)

QUARKUS-3290.md
rsvoboda: Should be `artifact => Maven artifacts`

QUARKUS-3290.md
rsvoboda: No text available that could be linked.

I'm not sure I should paste full trext into the test plan, that's why I linked the JIRA attachment.

```
Complete Corresponding Source Code
Complete Corresponding Source Code must include whatever is reasonably necessary for a package to be rebuilt
from source by a technically-skilled recipient with access to an appropriate build environment, without engaging in
excessive effort or experimentation. Items such as Makefiles, build and install scripts, build instructions, RPM spec
files, and POM files must be included with the source code as appropriate. Note that there are separate
requirements to make source code publicly available for Red Hat products for purposes of meeting our obligations
under the US and other export laws that may exceed or differ from the requirements described here. Please review
this link for more information on these separate requirements relating to export laws: separate requirements.
```

QUARKUS-3290.md
rsvoboda: Will update the PR ...

QUARKUS-3290.md
fedinskiy: Maybe we should check against binaries then? Checking first N symbols for non-UTF (non ascii?) content should be trivial enough and give us a reasonable protection.
With size check  would have both false negatives (small binaries) and false positives (huge source code files)

PR #153, opened 2023-11-28
Add test plan for QUARKUS-3535

michalvavrik:
### Links

JIRA: https://issues.redhat.com/browse/QUARKUS-3535

Quarkus documentation: https://quarkus.io/guides/all-config#quarkus-core_quarkus.thread-pool.queue-size

Native mode assumptions go down to the https://quarkus.io/guides/native-reference#multi-thread and the fact that I didn't really manage to reproduce it in native mode. Maybe it is safe to say that if I can't reproduce it here https://github.com/quarkus-qe/quarkus-test-suite/pull/1551 then users are less likely to be affected by any related issue.

I realize that Keycloak team had issues in OpenShift, but that really goes down to where big load happens, so it is not important whether you run this test on baremetal or OpenShift as there is no relevance of these changes to cloud environment.

   - Logging
   - Tracing
   - Metrics
   - Security
   - OpenAPI
   - Data sources
   - Frontend
   - Qute

### DIFF
--- a/QUARKUS-3535.md
+++ b/QUARKUS-3535.md
@@ -1,0 +1,34 @@
+# QUARKUS-3535 - Add a metric for rejected submissions to the pool in RHBQ 3.2
+
+JIRA: https://issues.redhat.com/browse/QUARKUS-3535
+
+PR: https://github.com/quarkusio/quarkus/pull/35541
+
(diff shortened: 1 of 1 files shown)


rsvoboda (approved):
LGTM. Thanks for deeper check of this.

PR #156, opened 2024-02-09
Adding a test plan for QUARKUS-3007

mjurc:
* Adding a test plan for QUARKUS-3007: Aarch64/ARM Support for both native and JVM Quarkus apps

### Links

JIRA: https://issues.redhat.com/browse/QUARKUS-3007

Quarkus documentation: https://access.redhat.com/articles/4966181

   - Logging
   - Tracing
   - Metrics
   - Security
   - OpenAPI
   - Data sources
   - Frontend
   - Qute

### DIFF
--- a/QUARKUS-3007.md
+++ b/QUARKUS-3007.md
@@ -1,0 +1,100 @@
+# QUARKUS-3007: Aarch64/ARM Support for both native and JVM Quarkus apps
+
+Jira: https://issues.redhat.com/browse/QUARKUS-3007
+
+The goal of this feature is to provide production support in the same scope for aarch64 architecture that RHBQ provides
+for x86-64 architecture.
+
+For the purposes of planning, the following tasks therefore need to be executed (ordered by priority):
+* Same test coverage and certification matrix on OCP on aarch64 for both JVM and native mode as on OCP on x86-64 (see
+  https://issues.redhat.com/browse/QUARKUS-3446)
+* Certification of RHBQ native mode on OCP on aarch64
+* Certification of RHBQ in both JVM and native mode in the bare metal scenarios on aarch64
+
+## Scope of the testing
+
+### General
+* Catch-up on the test coverage on OCP to be same for aarch64 as it is for x86-64
+  * Enabling Serverless tests in `openshift-arm` profile once RH Serverless supports aarch64
+  * Running test services that do not yet support aarch64 in a side OpenShift cluster running on x86-64
+    * This pertains to units to tens of tests across multiple modules 
+* Certification of RHBQ native mode on OCP on aarch64
+  * The full native OpenShift coverage will have to be ran in native mode with the Mandrel native builder container
+* Certification of both the JVM and native modes in bare metal scenarios will require all the aspects to be tested
+  * we support two JDK versions for JVM mode and one JDK version for native mode per release stream
+  * start-stop TS with RHBQ code starters
+  * start-stop special characters TS
+  * bare metal test suite
+  * Quarkus Quickstarts
+
+### Impact on test suites and testing automation
+* Catch-up on the test coverage on OCP to be same for aarch64 as it is for x86-64
+  * This means that the aarch64 OCP testing pipeline would require an additional x86-64 cluster to be deployed in AWS
+    and test services be prepared before the test execution by the test pipeline. The other part of this is that the
+    tests that currently use test services not supported on aarch64 would be moved to standalone test suite module, only
+    executed in `openshift-arm` profile
+  * The other aspect is testing on EUS OpenShift. A job executing the OpenShift interoperability modules on a matrix of 
+    JDK version and OpenShift version should be implemented.
+  * We will also need to implement periodic runs of OpenShift test suite in JVM mode with Quarkus built from main branch
+    on aarch64. 
+* Certification of RHBQ native mode on OCP on aarch64
+  * We need machines that execute the OpenShift test suite to be aarch64 too, as we need to compile the native binaries
+    on the same architecture they're deployed to.
+  * Pipeline will need to include reservation of aarch64 systems from Beaker, their provisioning and addition to
+    Jenkins besides the execution of the test coverage.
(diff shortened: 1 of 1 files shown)


michalvavrik:
Sorry, tough week, I'll look on Monday.

QUARKUS-3007.md
```diff
@@ -0,0 +1,100 @@
...
+  * Running test services that do not yet support aarch64 in a side OpenShift cluster running on x86-64
+    * This pertains to units to tens of tests across multiple modules 
+* Certification of RHBQ native mode on OCP on aarch64
+  * The full native OpenShift coverage will have to be ran in native mode with the Mandrel native builder container.
+* Certification of both the JVM and native modes in bare metal scenarios will require all the aspects to be tested
+  * we support two JDK versions for JVM mode and one JDK version for native mode per release stream
+  * start-stop TS with RHBQ code starters
+  * start-stop special characters TS
+  * bare metal test suite
+  * Quarkus quickstarts
```
michalvavrik, line 28:
```suggestion
  * Quarkus QuickStarts
```

QUARKUS-3007.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Impact on test suites and testing automation
+* Catch-up on the test coverage on OCP to be same for aarch64 as it is for x86-64
+  * This means that the aarch64 OCP testing pipeline would require an additional x86-64 cluster to be deployed in AWS
+    and test services be prepared before the test execution by the test pipeline. The other part of this is that the
+    tests that currently use test services not supported on aarch64 would be moved to standalone test suite module, only
```
michalvavrik, line 34: This can be big change depending on number of such a miscellaneous tests in one module. Can you give me a very rough idea how many such a tests from which modules it will be? If they are all with the `ts.arm.missing.services.excludes` it's going to be mess.

QUARKUS-3007.md
```diff
@@ -0,0 +1,100 @@
...
+    * 1x test service cluster - 6x xlarge x86-64 machines in AWS
+* Certification of both the JVM and native modes in bare metal scenarios on aarch64
+  * `startstop-ts-code-quarkus` - 1x large aarch64 machine from Beaker per tested JDK, minutes to tens of minutes 
+    execution time
+  * `startstop-ts-special-chars` - 1x large aarch64 machine from Beakerper tested JDK, minutes to tens of minutes 
+    execution time
+  * bare metal test suite
+    * JVM mode - 8x large aarch64 machine from Beaker per tested JDK, 1-2 hours execution time
+    * native mode - 8x xlarge aarch64 machine from Beaker, 2 hours execution time
+  * Quarkus quickstarts
```
michalvavrik, line 84:
```suggestion
  * Quarkus QuickStarts
```

QUARKUS-3007.md
mjurc: I'm tracking the task in https://issues.redhat.com/browse/QQE-258 - there's currently three test services we don't have on aarch64:
* RH AMQ streams - excluded by https://github.com/quarkus-qe/quarkus-test-suite/pull/1148/commits/36aab6dc891603cf36ee546274b38f037f19adb5 - 7 test cases in messaging module
* RH AMQ streams operator - excluded by https://github.com/quarkus-qe/quarkus-test-suite/pull/1148/commits/99feea23f4d23fb184c65ded7259963c6ac2b350 - 1 test case in messaging module
* RH SSO - https://github.com/quarkus-qe/quarkus-test-suite/pull/1148/commits/b995d9fe5c7c1616b6303d0f884e321371e91e14 - 10 test cases in security module

mjurc:
> LGTM. I have different personal preference - I think it is unnecessary to do moving to a separate module that only runs when `aarch64` is not used because:
> 
>     * all our tests are executed with the Quarkus QE Test Framework
> 
>     * we are able to detect architecture for example with the `uname -a` on Linux
> 
>     * you can't expect that test that does not support `aarch64` is run on that platform, not  sure why we just don't skip them automatically
> 
>     * you can keep profile and activate it explicitly if you like it better than FW (which someone will call magic, I call it a feature)
> 
> 
> However I can live with it...

I think I would prefer a better strategy than this too, but I don't expect the exclusions to expand - arm after all is part of wider initiative.

About detecting platform architecture from framework, I am all into this but note that the architecture of the test executors may be different from the architecture of the OCP clusters where the tests are ran. I suppose there's also way to detect that from the framework too, but I would hope that the exclusions and exclusions module is a stopgap measure that allows us to execute the coverage without changing the architecture of framework too much.

QUARKUS-3007.md
michalvavrik: that's acceptable amount, thanks

michalvavrik (approved):
LGTM

michalvavrik:
I suppose @rsvoboda won't review due to time limitations, so I'm going to merge this.

rsvoboda:
I can do after merge review :)

rsvoboda:
@mjurc I have only one comment: section Impact on resources - Product testing. it's not clear if there is implication on testing cycle length or our ability to process other streams when testing CR builds of feature releases (which include extended platforms)

mjurc:
> @mjurc I have only one comment: section Impact on resources - Product testing. it's not clear if there is implication on testing cycle length or our ability to process other streams when testing CR builds of feature releases (which include extended platforms)

Short answer is: I don't know.

Long answer:
* For machine time, I hope parallelism in execution can be established. The spin up of beaker machines is not that long in my experiments, but I don't have any signficantly large sample size to say how it will look in long term.
* For people/investigation time, it really depends on how different JDK and test services act on aarch64. So far, there weren't significant functional bugs on aarch64 - in fact, all the issues we've had was either with missing test services or missing productized dependencies. But I think native applications may be whole other matter.

I guess I can put those either in test plan or add a TODO to clarify once I know more.

rsvoboda:
I prefer adding these notes into TP with a note that we assume/hope this won't extend the needed time for testing cycle, but things can change ...

PR #161, opened 2024-03-13
QQE 378 HTTP module / REST use-cases - Development topic compression Brotli4j

jcarranzan:
### Links

Jira : [https://issues.redhat.com/browse/QQE-378](https://issues.redhat.com/browse/QQE-378)

Quarkus documentation:
- [Quarkus - http-reference#http-compression](https://quarkus.io/guides/http-reference#http-compression)
- [Quarkus - reactive-routes#http-compressio](https://quarkus.io/guides/reactive-routes#http-compression)
- [Vert.x - http_compression](https://vertx.io/docs/vertx-core/java/#_http_compression )
- [Vert.x - Handling compressed body](https://vertx.io/docs/vertx-core/java/#_handling_compressed_body)
- [Vert.x - http_compression_algorithms](https://vertx.io/docs/vertx-core/java/#_http_compression_algorithms)

   - Logging
   - Tracing
   - Metrics
   - Security
   - OpenAPI
   - Data sources
   - Frontend
   - Qute

### DIFF
--- a/QQE-378.md
+++ b/QQE-378.md
@@ -1,0 +1,53 @@
+# QQE-378 - HTTP module / REST use-cases - Development topic compression Brotli4j
+## Links
+Jira : [QQE-378](https://issues.redhat.com/browse/QQE-378)
+
+PR:  TODO
+## Scope of the testing
+Within the Quarkus framework, we have the option to leverage various compression strategies, including Gzip, Snappy, and Brotli4j. 
+Notably, Brotli4J, being a part of Vert.x, is inherently supported by Quarkus.
+
+Given the absence of existing test coverage for Brotli4J within this context,
+we have identified a need to encompass this area within our testing scope.
+ - Test whether Brotli4J compression library functions correctly within Quarkus, particularly in the context of REST endpoints using the Quarkus REST extension.
+ - The test coverage will be for JVM, native mode and Openshift.
+ - To use Brotli4J in quarkus compression we will need:
+   * Utilize Vert.x http server options  to add Brotli4J compressor support.   
+   * Ensure that the server is configured to respect the Accept-Encoding header of `br` to automatically
+     compress the response body with Brotli4j compression and send it back to the client.
+### Test scenarios 
+
+- Compression Enabled, Response Compressed and Decompressed
+   *    Set up a Quarkus application with the Brotli4J compression library.
+   *    Enable quarkus http configuration compression `quarkus.http.enable-compression=true`.
+   *    The response is compressed if the Content-Type header is set and the value is a compressed media type as configured specification in the next point.
+   *    Specify the media types supported to compress `quarkus.http.compress-media-types`.  
+   *    Note that the REST extension also make it possible to enable/disable compression declaratively using the annotations `io.quarkus.vertx.http.Compressed` and `io.quarkus.vertx.http.UnCompressed`.
+   *    Add the Brotli compressor using `.addCompressor(io.netty.handler.codec.compression.StandardCompressionOptions.brotli())`.
+   *    Send a request to a REST endpoint with the `.header("Accept-Encoding", "br")` and verify that the response is compressed using Brotli.
+   *    Verify for decompression, that we can use `quarkus.http.enable-decompression` when is it enabled, vert.x will decompress the request’s body if it’s compressed.
+- Compression Disabled, Response Not Compressed (Uncompressed)
+    *   Disable quarkus http configuration compression `quarkus.http.enable-compression=false`.
+    *   Send a request to a REST endpoint and verify that the response is not compressed.
+    *   The resource method is never compressed if is annotated with `@io.quarkus.vertx.http.Uncompressed`.
+- JSON Payloads, Compression Works up to 8192 bytes
+    *   Create a JSON payload of varying sizes up to 8192 bytes.
+    *   Send a request to a REST endpoint with the JSON payload and verify that the response is compressed
+  using Brotli.
+    * Create a JSON payload of more of 8192 bytes and check the error.
+-  Different Media Types, Compression Applied Correctly 
+     *   Create a plain text payload.
+     *   Create an XML payload.
+     *   Send a request to a REST endpoint with each payload and verify that the response is compressed using Brotli4J.
+
+    
+
+## Links to the resources
+- [Quarkus - http-reference#http-compression](https://quarkus.io/guides/http-reference#http-compression)
+- [Quarkus - reactive-routes#http-compressio](https://quarkus.io/guides/reactive-routes#http-compression)
+- [Vert.x - http_compression](https://vertx.io/docs/vertx-core/java/#_http_compression )
+- [Vert.x - Handling compressed body](https://vertx.io/docs/vertx-core/java/#_handling_compressed_body)
+- [Vert.x - http_compression_algorithms](https://vertx.io/docs/vertx-core/java/#_http_compression_algorithms)
+
+## Contacts
+- Tester: Jose Carranza <jcarranz@redhat.com>
--- a/QQE-378.md
+++ b/QQE-378.md
@@ -2,7 +2,7 @@
 ## Links
 Jira : [QQE-378](https://issues.redhat.com/browse/QQE-378)
 
-PR:  TODO
+PR:  https://github.com/quarkus-qe/quarkus-test-suite/pull/1717
 ## Scope of the testing
 Within the Quarkus framework, we have the option to leverage various compression strategies, including Gzip, Snappy, and Brotli4j. 
 Notably, Brotli4J, being a part of Vert.x, is inherently supported by Quarkus.


QQE-378.md
```diff
@@ -0,0 +1,43 @@
...
+Jira : [https://issues.redhat.com/browse/QQE-378](https://issues.redhat.com/browse/QQE-378)
+
+PR:  TODO
+## Scope of the testing
+ - Integration of the Brotli4J compression library with Quarkus, particularly in the context of REST endpoints using RestEasy Reactive.
+ - To use Brotli4J compression we will need a custom Server configuration
+   * Utilize `io.vertx.core.http.HttpServerOptions` to enable compression support.
+   * Add the Brotli compressor using `.addCompressor(io.netty.handler.codec.compression.StandardCompressionOptions.brotli())`.
+   * Ensure that the server is configured to respect the Accept-Encoding header to automatically compress the response body with one of the supported compressions and send it back to the client.
+  ### Test scenarios
```
michalvavrik, line 12:
```suggestion
### Test scenarios 
```

QQE-378.md
```diff
@@ -0,0 +1,43 @@
+# QQE-378 - HTTP module / REST use-cases - Development topic compression Brotli4j
+## Links
+Jira : [https://issues.redhat.com/browse/QQE-378](https://issues.redhat.com/browse/QQE-378)
```
michalvavrik, line 3:
```suggestion
Jira : [QQE-378](https://issues.redhat.com/browse/QQE-378)
```
optional, just an idea...

QQE-378.md
```diff
@@ -0,0 +1,43 @@
...
+
+PR:  TODO
+## Scope of the testing
```
michalvavrik, line 6: I think your objection needs to be map whether it works in native (probably not, but we need to now the state)

QQE-378.md
```diff
@@ -0,0 +1,43 @@
...
+    *  Send a request to a REST endpoint and verify that the response is not compressed
+- JSON Payloads, Compression Works up to 8192 bytes
+    *   Create a JSON payload of varying sizes up to 8192 bytes
+    *   Send a request to a REST endpoint with the JSON payload and verify that the response is compressed
+  using Brotli.
+    * Create a JSON payload of more of 8192 bytes and check the error.
+-  Different Media Types, Compression Applied Correctly 
+     *   Create a plain text payload
+     *   Create an XML payload
+     *   Send a request to a REST endpoint with each payload and verify that the response is compressed using Brotli
```
michalvavrik, line 31:
```suggestion
     *   Send a request to a REST endpoint with each payload and verify that the response is compressed using Brotli4J
```

QQE-378.md
```diff
@@ -0,0 +1,43 @@
...
+   * Ensure that the server is configured to respect the Accept-Encoding header to automatically compress the response body with one of the supported compressions and send it back to the client.
+  ### Test scenarios 
+
+- Compression Enabled, Response Compressed
+   *    Set up a Quarkus application with the Brotli4J compression library
+   *    Configure the HTTP server to enable compression using io.vertx.core.http.HttpServerOptions
+   *    Add the Brotli compressor using `.addCompressor(io.netty.handler.codec.compression.StandardCompressionOptions.brotli())`
+   *    Send a request to a REST endpoint and verify that the response is compressed using Brotli
+- Compression Disabled, Response Not Compressed
+    *    Set up a Quarkus application with the Brotli4J compression library
```
michalvavrik, line 20:
```suggestion
    *  Set up a Quarkus application with the Brotli4J compression library
```

QQE-378.md
```diff
@@ -0,0 +1,43 @@
...
+ - To use Brotli4J compression we will need a custom Server configuration
+   * Utilize `io.vertx.core.http.HttpServerOptions` to enable compression support.
+   * Add the Brotli compressor using `.addCompressor(io.netty.handler.codec.compression.StandardCompressionOptions.brotli())`.
+   * Ensure that the server is configured to respect the Accept-Encoding header to automatically compress the response body with one of the supported compressions and send it back to the client.
+  ### Test scenarios 
+
+- Compression Enabled, Response Compressed
+   *    Set up a Quarkus application with the Brotli4J compression library
+   *    Configure the HTTP server to enable compression using io.vertx.core.http.HttpServerOptions
+   *    Add the Brotli compressor using `.addCompressor(io.netty.handler.codec.compression.StandardCompressionOptions.brotli())`
```
michalvavrik, line 17: Sorry, this is bit implementation specific, I have no idea what it means. Can you use words instead of some methods that I know nothing about to describe what is your intention behind using this compressor? In the TP please (not in the comment)

michalvavrik (commented):
I could go on, but it's a draft so I just left few notes. Cheers :-)

QQE-378.md
```diff
@@ -0,0 +1,44 @@
+# QQE-378 - HTTP module / REST use-cases - Development topic compression Brotli4j
+## Links
+Jira : [QQE-378](https://issues.redhat.com/browse/QQE-378)
+
+PR:  TODO
+## Scope of the testing
+ - Integration of the Brotli4J compression library with Quarkus, particularly in the context of REST endpoints using RestEasy Reactive.
```
michalvavrik, line 7:
```suggestion
 - Integration of the Brotli4J compression library with Quarkus, particularly in the context of REST endpoints using Quarkus REST.
```

QQE-378.md
```diff
@@ -0,0 +1,44 @@
...
+ - The test coverage needs to be defined for both JVM and native mode.
```
michalvavrik, line 8: > defined

I think you define it here, so you are not making a plan for a plan. IMO you should just define it here, not say it needs to be defined. Maybe wrong words?

QQE-378.md
```diff
@@ -0,0 +1,44 @@
...
+ - To use Brotli4J compression we will need a custom Server configuration
+   * Utilize Vert.x http server options  to enable compression support.
```
michalvavrik, line 10: You basically want to set decompression supported https://github.com/eclipse-vertx/vert.x/blob/master/src/main/asciidoc/http.adoc#handling-compressed-body but there [quarkus.http.enable-decompression](https://quarkus.io/guides/all-config#quarkus-vertx-http_quarkus-http-enable-decompression) and same for a management interface. I thin kyou should use them instead.

QQE-378.md
```diff
@@ -0,0 +1,44 @@
...
+   * Include the compressor using Brotli4J options compression.
+   * Ensure that the server is configured to respect the Accept-Encoding header to automatically compress the response body with one of the supported compressions and send it back to the client.
+### Test scenarios 
+
+- Compression Enabled, Response Compressed
+   *    Set up a Quarkus application with the Brotli4J compression library
+   *    Configure the HTTP server to enable compression using io.vertx.core.http.HttpServerOptions
```
michalvavrik, line 17: Same question as above, why is that necessary? Shouldn't you configure Quarkus instead?

QQE-378.md
```diff
@@ -0,0 +1,44 @@
...
+   *    Send a request to a REST endpoint and verify that the response is compressed using Brotli
+- Compression Disabled, Response Not Compressed
+    *   Set up a Quarkus application with the Brotli4J compression library
+    *   Configure the HTTP server to disable compression using io.vertx.core.http.HttpServerOptions
+    *   Send a request to a REST endpoint and verify that the response is not compressed
+- JSON Payloads, Compression Works up to 8192 bytes
+    *   Create a JSON payload of varying sizes up to 8192 bytes
+    *   Send a request to a REST endpoint with the JSON payload and verify that the response is compressed
+  using Brotli.
+    * Create a JSON payload of more of 8192 bytes and check the error.
```
michalvavrik, line 28: I don't understand why it shouldn't work, please add links and explanation into the TP. Here https://github.com/quarkusio/quarkus/issues/34719#issuecomment-1638142179 Clement suggested more than 8192 is supported when using Vert.x directly?

michalvavrik: AKA if this will only work when Quarkus bumps to next major Vert.x version in a year time, you should document it as well. Otherwise it's strange to just expect errors without an explanation.

QQE-378.md
jcarranzan: Well, quarkus options like `quarkus.http.enable-compression` and `quarkus.http.enable-decompression` works by default for gzip so for brotli4J we need both options enabled and use Vert.x HTTP server options in some way (as you mentioned above) or directly the Brotli4J library API.

QQE-378.md
```diff
@@ -0,0 +1,44 @@
...
+## Links
+Jira : [QQE-378](https://issues.redhat.com/browse/QQE-378)
+
+PR:  TODO
+## Scope of the testing
+ - Integration of the Brotli4J compression library with Quarkus, particularly in the context of REST endpoints using RestEasy Reactive.
+ - The test coverage needs to be defined for both JVM and native mode.
+ - To use Brotli4J compression we will need a custom Server configuration
+   * Utilize Vert.x http server options  to enable compression support.
+   * Include the compressor using Brotli4J options compression.
```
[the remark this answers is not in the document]
jcarranzan: I answered it above, and here you are right I will re-write those points to provide clarification.

QQE-378.md
jcarranzan: Clement wrote, "Our Brotli support only works up to 8192 (1024*8) bytes." 
up to = until= not exceeding

QQE-378.md
jcarranzan: As far as I know, Quarkus itself is not capable of managing brotli4J but yes gzip. FOr that reason, vert.x configuration HTTP server options becomes relevant in this context.

QQE-378.md
michalvavrik: ok, but Quarkus relies on Vert.x that doesn't limit that. Saying 8192 is our number and we shall test it is crazy. It must be documented, there must be opened issue and linked to this TP. Thanks

michalvavrik: It's not directed to you, but I just don't want to accept this.

michalvavrik:
Hello @jcarranzan , I'm not going to continue discussion in individual comment threads because there seems to be something I don't link about this TP in general, not particular lines. You say:

> As far as I know, Quarkus itself is not capable of managing brotli4J but yes gzip. For that reason, vert.x configuration HTTP server options become relevant in this context.

I don't know where these information comes form, but it doesn't seem to be correct.

> Well, quarkus options like quarkus.http.enable-compression and quarkus.http.enable-decompression works by default for gzip so for brotli4J we need both options enabled and use Vert.x HTTP server options in some way (as you mentioned above) or directly the Brotli4J library API.

I gave you link for the `quarkus.http.enable-compression ` already:  https://quarkus.io/guides/all-config#quarkus-vertx-http_quarkus-http-enable-compression

and it says: _Note that the compression format (e.g., gzip) must be specified in the Content-Encoding header in the request._ which means you just set content-encoding to the `br` and that is it. 

Please consider rewriting this TP so that we actually test Quarkus support to Brotli4j compression and not Vert.x support. I think we should use Quarkus constructs, that is we set these configuration properties (enable compression) we use annotations and headers. If that is not sufficient, then please consider opening bug in Quarkus. Thank you

BTW when looking into this, I also mentioned netty extensions handles brotli4j already, so you don't need to add it and it should work in native.

QQE-378.md
michalvavrik: Point is, some PR comment does not determine contract.

michalvavrik: But you are right, I read Clement's comment wrong. But I don't like that comment as we need documentation for users.

jcarranzan:
Hi @michalvavrik , my comments about this PR came from the fact that last week I was experimenting with the code and the documentation. Indeed, there isn't much information about Brotli4J compression in the Quarkus docs, and I couldn't find much in the Quarkus core code either, except that the Netty extensions handle Brotli4J. Therefore, in Vert.x, it's also handled with custom HTTP server configuration, as I did a draft PR here ([draft)](https://github.com/quarkus-qe/quarkus-test-suite/pull/1717).

That being said, I pushed some classes with the basic Quarkus setup to my branch to see if I could implement Brotli4J compression, as you suggested, but I wasn't successful. So, I would appreciate it if you could take a look and see if you notice anything that I might be missing. (branch--> https://github.com/jcarranzan/quarkus-test-suite/tree/brotli4j-coverage-3.8). Thanks

michalvavrik:
> Hi @michalvavrik , my comments about this PR came from the fact that last week I was experimenting with the code and the documentation. Indeed, there isn't much information about Brotli4J compression in the Quarkus docs, and I couldn't find much in the Quarkus core code either, except that the Netty extensions handle Brotli4J. Therefore, in Vert.x, it's also handled with custom HTTP server configuration, as I did a draft PR here (https://github.com/quarkus-qe/quarkus-test-suite/pull/1717.

Tried it. Using `io.vertx.core.http.HttpServerOptions#addCompressor` seems only way. The `Content-Encoding` is only for decompression, my fault. The `Accept-Encoding` can only use to signal client support for existing compressors.

So to conclude this TP, I suggest:

- Can you add to the test plan choosing between gzip and br by that header if that makes sense to you.
- Consider creating documentation issue in the Quarkus that shows how to add support for other than default compressor? I re-read all the related docs and I don't think it's user friendly. Choice is yours though.
- We already have compressed / decompressed annotations tested somewhere, can you also add it to the TP for Brotli4j?
- Explain in the TP why up to `8192` magical number is tested. I stick to my previous comments, some PR header does not establish a contract and we shouldn't follow it. What users need to know should be found in the docs.

Please answer following questions:

- https://issues.redhat.com/browse/QQE-378 only speaks about compression, isn't it strange to test compression support for Brotli4j and not decompression?
- https://issues.redhat.com/browse/QQE-378 speaks about other compressors like Snappy, TP should explain why chose to test Brotli4j and not others.
- Why isn't OpenShift relevant for this testing?

QQE-378.md
```diff
@@ -0,0 +1,45 @@
...
+   * Utilize Vert.x http server options  to add Brotli4J compressor support.   
+   * Ensure that the server is configured to respect the Accept-Encoding header of `br` to automatically
+     compress the response body with brotli4j compression and send it back to the client.
+### Test scenarios 
+
+- Compression Enabled, Response Compressed
+   *    Set up a Quarkus application with the Brotli4J compression library
+   *    Enable quarkus http configuration compression `quarkus.http.enable-compression=true`
+   *    Specify the media types supported to compress `quarkus.http.compress-media-types`
+   *    Configure the HTTP server to enable compression using io.vertx.core.http.HttpServerOptions.
```
michalvavrik, line 19: This step should not be necessary.

michalvavrik:
In regards of the 8192, can you please check whether raising https://quarkus.io/guides/all-config#quarkus-vertx-http_quarkus-http-limits-max-chunk-size helps? If not, there should be opened documentation issue linked with this TP. Thank you

QQE-378.md
```diff
@@ -0,0 +1,52 @@
+# QQE-378 - HTTP module / REST use-cases - Development topic compression Brotli4j
+## Links
+Jira : [QQE-378](https://issues.redhat.com/browse/QQE-378)
+
+PR:  TODO
```
michalvavrik, line 5: Can TODO already be replaced with the https://github.com/quarkus-qe/quarkus-test-suite/pull/1717 or are there are PRs connected to this TP? If not, we can probably make this TP ready for review?

jcarranzan:
> Please answer following questions:
> 
>     * https://issues.redhat.com/browse/QQE-378 only speaks about compression, isn't it strange to test compression support for Brotli4j and not decompression? Good point, now I've added some lines related to decompression also.
> 
>     * https://issues.redhat.com/browse/QQE-378 speaks about other compressors like Snappy, TP should explain why chose to test Brotli4j and not others. Added it also.
> 
>     * Why isn't OpenShift relevant for this testing? Yes, it's, added.
Thanks.

michalvavrik (approved):

PR #171, opened 2024-04-17
Adding a test plan for QUARKUS-3461

gtroitsk:
Adding a test plan for QUARKUS-3461: verify behaviour and functionality of Quarkus CLI update command with RHBQ

### Links

JIRA:  https://issues.redhat.com/browse/QUARKUS-3461

Quarkus documentation:
https://quarkus.io/guides/update-quarkus

   - Logging
   - Tracing
   - Metrics
   - Security
   - OpenAPI
   - Data sources
   - Frontend
   - Qute

### DIFF
--- a/QUARKUS-3461.md
+++ b/QUARKUS-3461.md
@@ -1,0 +1,41 @@
+# QUARKUS-3461 Verify behaviour and functionality of Quarkus CLI update command with RHBQ
+- JIRA link: https://issues.redhat.com/browse/QUARKUS-3461
+- Quarkus guides:
+    - https://quarkus.io/guides/update-quarkus
+
+## Scope of the testing
+- Exploratory testing of the feature over supported versions of RHBQ and Quarkus upstream
+- Writing test coverage for this feature is continual work, requires update for every release of Quarkus
+- Test update between each of Quarkus LTS release versions
+- Test update of multi-module projects
+- Test coverage in QQE TS:
+  - Test most frequently applied OpenRewrite recipes:
+    - org.openrewrite.properties.ChangePropertyKey
+    - org.openrewrite.quarkus.ChangeQuarkusPropertyKey
+    - org.openrewrite.java.ChangeMethodName
+    - org.openrewrite.java.dependencies.ChangeDependency
+    - org.openrewrite.maven.UpgradePluginVersion
+    - org.openrewrite.maven.UpgradeDependencyVersion
+    - org.openrewrite.quarkus.AddQuarkusProperty
+    - org.openrewrite.quarkus.DeleteQuarkusProperty
+    - org.openrewrite.maven.AddDependency
+  - Dynamically create Quarkus applications and manually copy test classes from test classpath. Each application should contain a group of dependencies and properties that were modified in the [recipes](https://github.com/quarkusio/quarkus-updates/tree/main/recipes/src/main/resources/quarkus-updates/core). Multi-module projects can only be created manually
+  - Created apps should reflect the customer's usage scenario including test coverage  
+  - Test that initial application was successfully built. Update application to selected stream and verify that app was successfully run, test expected changes based on [recipes](https://github.com/quarkusio/quarkus-updates/tree/main/recipes/src/main/resources/quarkus-updates/core) and [Quarkus migration guides](https://github.com/quarkusio/quarkus/wiki/Migration-Guides)
+
+## Getting familiar with the feature
+Following actions were taken to ensure familiarity:
+- Ensure documentation provides clear explanation on configuration options
+- Ensure good user experience and simplicity of use
+
+## Existing test coverage
+- [Quarkus updates](https://github.com/quarkusio/quarkus-updates/tree/main/recipes-tests/src/test) 
+
+### Impact on test suites and testing automation
+- New tests will be added to [quarkus-cli](https://github.com/quarkus-qe/quarkus-test-suite/tree/main/quarkus-cli) module in QQE TS.
+
+### Impact on resources
+- Tests in QE Test Suite increase execution time up to 10 minutes in JVM. In case of testing application in native mode in future, this will extend execution time too.
+
+## Contacts
+- Tester: Georgii Troitskii <gtroitsk@redhat.com>


QUARKUS-3461.md
```diff
@@ -0,0 +1,28 @@
...
+- JIRA link: https://issues.redhat.com/browse/QUARKUS-3461
+- Quarkus Guides:
+    - https://quarkus.io/guides/update-quarkus
+
+## Scope of the testing
+- Exploratory testing of the feature over Quarkus upstream and last released version of RHBQ
+- Verify that the `quarkus update` command explicitly warns users about deprecated/unsupported extensions during migration to the newer version.
+  In most cases deprecated extensions are moved to `io.quarkiverse`, which is accepted for Quarkus upstream but not for RHBQ.
+- Test coverage in QQE TS:
+  - Create a Quarkus/RHBQ app with the oldest possible stream and update to the latest version - verify new Quarkus platform version and app functionality
```
michalvavrik, line 11: This doesn't tell me what are you going to test it because `quarkus update` will only have effect if some of recipes are applicable. If you meant to prepare the old version of app in a way that it needs to be updated, please be more specific in the plan.

QUARKUS-3461.md
```diff
@@ -0,0 +1,28 @@
+# QUARKUS-3461 Verify behaviour and functionality of Quarkus CLI update command with RHBQ
+- JIRA link: https://issues.redhat.com/browse/QUARKUS-3461
+- Quarkus Guides:
+    - https://quarkus.io/guides/update-quarkus
+
+## Scope of the testing
+- Exploratory testing of the feature over Quarkus upstream and last released version of RHBQ
+- Verify that the `quarkus update` command explicitly warns users about deprecated/unsupported extensions during migration to the newer version.
+  In most cases deprecated extensions are moved to `io.quarkiverse`, which is accepted for Quarkus upstream but not for RHBQ.
```
michalvavrik, line 9: How are you going to automate that? I don't think there is a point in testing past releases unless the used recipe is reused and if it is, then you should test the new change, not the old one because there always can be a typo.

QUARKUS-3461.md
```diff
@@ -0,0 +1,28 @@
...
+- Test coverage in QQE TS:
+  - Create a Quarkus/RHBQ app with the oldest possible stream and update to the latest version - verify new Quarkus platform version and app functionality
+
+## Getting familiar with the feature
+Following actions were taken to ensure familiarity:
+- Ensure documentation provides clear explanation on configuration options
+- Ensure good user experience and simplicity of use
+
+## Existing test coverage
+- [Quarkus updates](https://github.com/quarkusio/quarkus-updates/tree/main/recipes-tests/src/test)
```
michalvavrik, line 19: Camel Quarkus team at the very least explored this in past, I think they had some requirement to test this command, not sure if they succeeded. Did you checked for their test coverage or contacted them? I think it would be interesting to know what they are doing and how.

michalvavrik (commented):
Truth be told, I was afraid when I saw the ticket and it fallen on you :-) So:

- TP is alright except my comment below
- I might be a troublemaker here, IMO you might want to change reviewer to @rsvoboda or @mjurc ; absolutely fine with me


Ad scope:

- as I understand TP, this TD is not actually going to test the feature
- I think to create proper TD will be a very hard work

Here is what I'd like you to do:

- create a test for hardcoded list of recipes that are most often applied, for example (please create your own list):
  - `org.openrewrite.maven.UpgradeDependencyVersion`
  - `org.openrewrite.java.ChangePackage`
  - `org.openrewrite.java.dependencies.ChangeDependency`
  - `org.openrewrite.java.ChangeType`
  - `org.openrewrite.properties.ChangePropertyKey`
  - `org.openrewrite.text.FindAndReplace`
- previous version you are going to create should have prepared a state that could be updated and now, that is a huge problem:
  - for extension name changes, you can rely on code starts so that we can assert that application after update is compiled and started
  - Java change types you would have to analyze by Reflection API and declare them when Class type as `ChangedClass changedClass = null;` and annotation must be on something based on the class etc. Why is that important? You know that recipes work because they are reused, but you have no idea whether there is no typo so that new Java type is not even able to compile.
  - for `org.openrewrite.properties.ChangePropertyKey` you cannot test well that change was semantically correct,  but you can test that new property is known to Quarkus (Quarkus is capable to report both runtime and buildtime uknown properties, so if he reports newly set property, somethings wrong :-)) but you will need to be able to handle application failure as setting random config properties will sooner or later fail startup
- we need to create new test coverage for every release based on the recipes in place because going with automation far would be ridiculous, so facilitate the process
- make it clear in TP that TD for this feature is never ever finished, it must be continual work, no genius can write generic test for this

Here is how the workflow should look like:

- we need to update TD for every release, so on push https://github.com/quarkusio/quarkus-updates we need to have opened issue in our TS or something like that will be annoying enough for us to mention that we need to update test coverage; or on each Quarkus release; anyway, it must be annoying so that we pay attention
- companies often use LTS, we should keep update between next LTS: 3.2 -> 3.8, 3.8 -> [future version]
- tests run during RHBQ testing should test update from previous RHBQ release to the current one

The most sensible thing is IMHO to ignore all suggested automation creation of new recipes, instead, create utils and method to facilitate testing of dependencies, properties etc. and only thing that human will need to do between releases is to:

1. human check what changed in recipes
2. human creates sensible application based on templates you prepare, so he will say:
  - here are application properties
  - here are dependencies before
  - assert that when run, you can detect A, B, C
3. drop obsolete test coverage for unsupported releases

rsvoboda:
@gtroitsk can you please update this doc? We discussed few things last week.

gtroitsk:
> @gtroitsk can you please update this doc? We discussed few things last week.

Sure, I will update it later today

QUARKUS-3461.md
```diff
@@ -0,0 +1,40 @@
...
+    - org.openrewrite.properties.ChangePropertyKey
+    - org.openrewrite.quarkus.ChangeQuarkusPropertyKey
+    - org.openrewrite.java.ChangeMethodName
+    - org.openrewrite.java.dependencies.ChangeDependency
+    - org.openrewrite.maven.UpgradePluginVersion
+    - org.openrewrite.maven.UpgradeDependencyVersion
+    - org.openrewrite.quarkus.AddQuarkusProperty
+    - org.openrewrite.quarkus.DeleteQuarkusProperty
+    - org.openrewrite.maven.AddDependency
+  - Add static Quarkus applications, where each application should contain a group of dependencies and properties that were modified in the [recipes](https://github.com/quarkusio/quarkus-updates/tree/main/recipes/src/main/resources/quarkus-updates/core)
```
michalvavrik, line 21: Did you abandon dynamically created apps we talked about with the Quarkus CLI service? I am not sure if having full projects is better than something like: `cliClient.createApplication("app", new Extensions { QUARKUS_REST, SECURITY}, new Class[] { TestClass1.class, TestClass2.class }, "quarkus.oidc.enabled=false\n")`.

michalvavrik (commented):
looks good, almost there

QUARKUS-3461.md
```diff
@@ -0,0 +1,40 @@
+# QUARKUS-3461 Verify behaviour and functionality of Quarkus CLI update command with RHBQ
+- JIRA link: https://issues.redhat.com/browse/QUARKUS-3461
+- Quarkus Guides:
```
michalvavrik, line 3:
```suggestion
- Quarkus guides:
```

QUARKUS-3461.md
```diff
@@ -0,0 +1,40 @@
...
+- Ensure good user experience and simplicity of use
+
+## Existing test coverage
+- [Quarkus updates](https://github.com/quarkusio/quarkus-updates/tree/main/recipes-tests/src/test) 
+
+### Impact on test suites and testing automation
+- New tests will be added inside [quarkus-cli](https://github.com/quarkus-qe/quarkus-test-suite/tree/main/quarkus-cli) module in QQE TS.
+
+### Impact on resources
+- Tests in QE Test Suite increase execution time up to 10 minutes in JVM
```
michalvavrik, line 37: 10 minutes in JVM sounds ok if that's necessary (will depend on required tests), but: in the future there might be need to DEV  and native test, so you might mention it.

Optionally this TP should contain goal to remove tests for unsupported streams with each addition.

QUARKUS-3461.md
gtroitsk: For simple apps Quarkus CLI service is enough, but multi-module project cannot be created using CLI

michalvavrik: I can't think of an example of a test scenario where multi-module project is required and we are going to create it. I can see you addressed my comment so closing this thread.

QUARKUS-3461.md
```diff
@@ -0,0 +1,41 @@
...
+## Getting familiar with the feature
+Following actions were taken to ensure familiarity:
+- Ensure documentation provides clear explanation on configuration options
+- Ensure good user experience and simplicity of use
+
+## Existing test coverage
+- [Quarkus updates](https://github.com/quarkusio/quarkus-updates/tree/main/recipes-tests/src/test) 
+
+### Impact on test suites and testing automation
+- New tests will be added inside [quarkus-cli](https://github.com/quarkus-qe/quarkus-test-suite/tree/main/quarkus-cli) module in QQE TS.
```
michalvavrik, line 35:
```suggestion
- New tests will be added to [quarkus-cli](https://github.com/quarkus-qe/quarkus-test-suite/tree/main/quarkus-cli) module in QQE TS.
```

QUARKUS-3461.md
```diff
@@ -0,0 +1,41 @@
+# QUARKUS-3461 Verify behaviour and functionality of Quarkus CLI update command with RHBQ
+- JIRA link: https://issues.redhat.com/browse/QUARKUS-3461
+- Quarkus guides:
+    - https://quarkus.io/guides/update-quarkus
+
+## Scope of the testing
+- Exploratory testing of the feature over supported versions of RHBQ and Quarkus upstream
+- Writing test coverage for this feature is continual work, requires update for every release of Quarkus
+- Test update between each of Quarkus LTS release versions
+- Test update of multi-module projects
```
michalvavrik, line 10: Quick check, is this even possible right now? I remember trying it on TS when javax -> jakarta was required and it didn't work. I cannot see it https://quarkus.io/guides/update-quarkus. It's ok as long as it is expected to work.

QUARKUS-3461.md
```diff
@@ -0,0 +1,41 @@
...
+    - org.openrewrite.properties.ChangePropertyKey
+    - org.openrewrite.quarkus.ChangeQuarkusPropertyKey
+    - org.openrewrite.java.ChangeMethodName
+    - org.openrewrite.java.dependencies.ChangeDependency
+    - org.openrewrite.maven.UpgradePluginVersion
+    - org.openrewrite.maven.UpgradeDependencyVersion
+    - org.openrewrite.quarkus.AddQuarkusProperty
+    - org.openrewrite.quarkus.DeleteQuarkusProperty
+    - org.openrewrite.maven.AddDependency
+  - Dynamically create Quarkus applications and manually copy tests from test classpath. Each application should contain a group of dependencies and properties that were modified in the [recipes](https://github.com/quarkusio/quarkus-updates/tree/main/recipes/src/main/resources/quarkus-updates/core). Multi-module projects can only be created manually
```
michalvavrik, line 22:
```suggestion
  - Dynamically create Quarkus applications and manually copy test classes from test classpath. Each application should contain a group of dependencies and properties that were modified in the [recipes](https://github.com/quarkusio/quarkus-updates/tree/main/recipes/src/main/resources/quarkus-updates/core). Multi-module projects can only be created manually
```

michalvavrik (commented):
lgtm; I left very minor comments. I suggest to wait a day or two if @mjurc or @rsvoboda have what to say, then we can merge it

QUARKUS-3461.md
michalvavrik: if you think `inside` is correct, let's keep it

QUARKUS-3461.md
gtroitsk: I think it's not possible now, we need to push dev to support this feature. I think, test update of complex RHBQ applications is necessary. Last time we talk with @rsvoboda, he mentioned that we need test complex applications

michalvavrik: thanks, let's keep it then

michalvavrik (approved):

PR #177, opened 2024-06-11
Add test plan for QUARKUS-4430 | http metrics should provide a path instead of REDIRECTION and NOT_FOUND

jcarranzan:
### Links

JIRA:
https://issues.redhat.com/browse/QUARKUS-4430

   - Logging
   - Tracing
   - Metrics
   - Security
   - OpenAPI
   - Data sources
   - Frontend
   - Qute

### DIFF
--- a/QUARKUS-4430.md
+++ b/QUARKUS-4430.md
@@ -1,0 +1,37 @@
+# QUARKUS-3535 - Add a metric for rejected submissions to the pool in RHBQ 3.2
+
+JIRA: https://issues.redhat.com/browse/QUARKUS-4430
+
+PR: https://github.com/quarkusio/quarkus/pull/39583
+
+Upstream issue: https://github.com/quarkusio/quarkus/issues/39581
+
+Verify that URIs in metrics tags are correctly recorded when they match a client or server pattern,
+especially in cases of redirection (302) and not found (404) responses.
+
+## Scope of the testing
+* Run Quarkus micrometer Prometheus application and verify metric tags for:
+   * Redirections (HTTP 302)
+   * Not Found responses (HTTP 404)
+   * Empty path responses (HTTP 204)
+   * Dynamic path segments (HTTP 301)
+
+* The test coverage will be verified for JVM and Native.
+
+## Existing test coverage
+* We have already tests related to monitoring metrics with Prometheus but not specific test
+  about this improvement feature related to tag redirections and not found responses recorded in metrics.
+ 
+
+### Impact on test suites and testing automation
+* New test class will be added into [monitoring](https://github.com/quarkus-qe/quarkus-test-suite/tree/main/monitoring/micrometer-prometheus) 
+
+### Impact on resources
+* One baremetal test execution in JVM mode and native mode.
+* Overall QE Test Suite execution time will be greater roughly by a few seconds
+
+## Contacts
+* Tester: Jose Carranza <jcarranz@redhat.com>
+
+## References
+- [http metrics should provide a path instead of REDIRECTION and NOT_FOUND](https://github.com/quarkusio/quarkus/issues/39581)


QUARKUS-4430.md
```diff
@@ -0,0 +1,35 @@
...
+* We have already tests related to monitoring metrics with Prometheus and OpenId Connect Keycloak but not specific test
+  about this improvement feature related to tag redirections and not found responses.
+ 
+
+### Impact on test suites and testing automation
+* New test class will be added into [monitoring](https://github.com/quarkus-qe/quarkus-test-suite/tree/main/monitoring/micrometer-prometheus-oidc) 
+
+### Impact on resources
+* One baremetal test execution in JVM mode
+* Test require any KeyCloak service container
```
michalvavrik, line 28:
```suggestion
* Test require any Keycloak service container
```
I don't understand why it would require Keycloak at all, can you elaborate?

QUARKUS-4430.md
```diff
@@ -0,0 +1,35 @@
...
+## Existing test coverage
+* We have already tests related to monitoring metrics with Prometheus and OpenId Connect Keycloak but not specific test
+  about this improvement feature related to tag redirections and not found responses.
+ 
+
+### Impact on test suites and testing automation
+* New test class will be added into [monitoring](https://github.com/quarkus-qe/quarkus-test-suite/tree/main/monitoring/micrometer-prometheus-oidc) 
+
+### Impact on resources
+* One baremetal test execution in JVM mode
```
michalvavrik, line 27: we generally test everything in both JVM and native mode unless we have a reason for the opposite; now, the Keycloak only runs in JVM IIRC but we don't care about Keycloak per say, just about Quarkus

QUARKUS-4430.md
```diff
@@ -0,0 +1,35 @@
...
+* Test require any KeyCloak service container
+* Overall QE Test Suite execution time will be greater roughly by 10 seconds
```
michalvavrik, line 29: that can only be true if you integrate new test coverage into existing test class, can you do that? if so, cool

QUARKUS-4430.md
```diff
@@ -0,0 +1,35 @@
+# QUARKUS-3535 - Add a metric for rejected submissions to the pool in RHBQ 3.2
+
+JIRA: https://issues.redhat.com/browse/QUARKUS-4430
+
+PR: https://github.com/quarkusio/quarkus/pull/39583
+
+Upstream issue: https://github.com/quarkusio/quarkus/issues/39581
+
+Verify that URIs in metrics tags are correctly recorded when they match a client or server pattern,
```
michalvavrik, line 9: my understanding is that this change is about path templates, so having test coverage for 302 and 404 is cool, but please stress in the TP what is it that you want to test, e.g. `/moved/{id}`. because you need to create a test coverage that pass with 3.8.5 and but fails with 3.8.4. thanks

now, I understand that this text comes from the PR title `Keep the URIs in the metrics tag if they match a client or server pattern`, but I am not convinced it's correct term. It is path pattern in either client or server path.

QUARKUS-4430.md
```diff
@@ -0,0 +1,35 @@
...
+especially in cases of redirection (302) and not found (404) responses.
+
+## Scope of the testing
+* Run Quarkus micrometer Prometheus application with OpenId Connect Keycloak and verify in case of redirections (302)
```
michalvavrik, line 13: _OpenId Connect Keycloak_ is not really relevant for scope of testing, it's relevant for a motivation of this change

QUARKUS-4430.md
```diff
@@ -0,0 +1,35 @@
...
+  and not found (404) response the Uris metrics tags are correctly recorded.
+* The test coverage will be verified for JVM.
```
michalvavrik, line 15: judging by code changes this will work in native, but if you integrate this into existing test class, you will run it in native anyway (as you should)

QUARKUS-4430.md
```diff
@@ -0,0 +1,35 @@
...
+JIRA: https://issues.redhat.com/browse/QUARKUS-4430
+
+PR: https://github.com/quarkusio/quarkus/pull/39583
+
+Upstream issue: https://github.com/quarkusio/quarkus/issues/39581
+
+Verify that URIs in metrics tags are correctly recorded when they match a client or server pattern,
+especially in cases of redirection (302) and not found (404) responses.
+
+## Scope of the testing
```
michalvavrik, line 12: I'd like you to add test cases for paths with empty path segments; the https://github.com/quarkusio/quarkus/pull/39583/files is comparing paths and we should be sure that it is using normalized paths (which I suppose it does, but let's have tests). e.g. `/moved////{id}`

michalvavrik (commented):
TP looks fine, but I still found a way to complain; please see comments below

QUARKUS-4430.md
jcarranzan: Well, this issue development has been done for a Keycloak request, you can see here--> https://issues.redhat.com/browse/QUARKUS-4430 
_"Keycloak wants to raise a request to backport the enhancement for reflecting redirected URLs on HTTP Metrics. The Quarkus version in question is 3.8.

The relevant PR to be backported: https://github.com/quarkusio/quarkus/pull/39583 "_

michalvavrik: Let's add it to the motivation (the description paragraph), but you cannot use Keycloak in this testing because you don't have a use case. If you have in mind how to use the Keycloak in relation to the backport, please show me.

QUARKUS-4430.md
jcarranzan: I'll try it.

QUARKUS-4430.md
jcarranzan: For that reason, my idea was create the tests into https://github.com/quarkus-qe/quarkus-test-suite/blob/main/monitoring/micrometer-prometheus-oidc/ that it's used Keycloak and prometheus metric, so in that case we will have an integration test checking those urls for the cases Not found and redirection....

jcarranzan: I've moved the test just  to /monitoring/micrometer-prometheus without keycloak because is not relevant here.

QUARKUS-4430.md
```diff
@@ -0,0 +1,37 @@
...
+* Run Quarkus micrometer Prometheus application and verify metric tags for:
+   * Redirections (HTTP 302)
+   *   Not Found responses (HTTP 404)
```
michalvavrik, line 15:
```suggestion
   * Not Found responses (HTTP 404)
```

QUARKUS-4430.md
```diff
@@ -0,0 +1,37 @@
...
+   *  Empty path responses (HTTP 204)
+   *  Dynamic path segments (HTTP 301)
```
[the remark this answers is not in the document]
michalvavrik: BTW why are they called dynamic path segments? I am just curious why you use this word for 301. I'm probably missing something.

michalvavrik (approved):

PR #179, opened 2024-07-02
Add test plan of JSON payloads in Quarkus REST endpoints

jcarranzan:
### Links

JIRA:
 [QQE-379](https://issues.redhat.com/browse/QQE-379)

Quarkus documentation:
- [Writing JSON REST Services ](https://quarkus.io/guides/rest-json)
- [Customizing the ObjectMapper in REST Client Jackson](https://quarkus.io/guides/rest-client#customizing-the-objectmapper-in-rest-client-jackson)
- [Creating the first JSON REST service](https://quarkus.io/guides/resteasy#creating-the-first-json-rest-service)

   - Logging
   - Tracing
   - Metrics
   - Security
   - OpenAPI
   - Data sources
   - Frontend
   - Qute

### DIFF
--- a/QQE-379.md
+++ b/QQE-379.md
@@ -1,0 +1,80 @@
+# QQE-379 - HTTP module / REST use-cases - Development topics - JSON payload
+## Links
+
+Jira : [QQE-379](https://issues.redhat.com/browse/QQE-379)
+
+TD: https://github.com/quarkus-qe/quarkus-test-suite/pull/1890
+
+This test plan outlines the testing for different JSON payloads in Quarkus REST endpoints.
+It aims to ensure the correct handling of JSON data in various scenarios, including successful processing, invalid payloads, error handling, and complex data structures.
+
+## Scope of the testing
+
+* The test coverage will be verified for environments:  JVM, native mode and Openshift.
+* Libraries used: quarkus-rest-jackson, quarkus-rest-jsonb
+* Payloads: valid, invalid, large, complex, missing/incorrect/empty/null data types, special characters.
+
+### What to test
+
+#### Successful JSON payload processing:
+Verify that Quarkus REST endpoints successfully process valid JSON payloads.
+  - **Payload:** simple json file. 
+  - **HTTP Method:** `POST`
+  - **Deserialization:** Ensure accurate conversion of JSON payloads, that it's performed using the configured JSON libraries (quarkus-rest-jackson, quarkus-rest-jsonb).
+  - **HTTP Status Code:** Expect `201 Created` upon successful payload processing.
+  - **Response Content:** Validate that the response contains the expected message.
+
+#### Check large JSON with complex structures payload processing:
+Verify that the application can efficiently process large JSON payloads containing nested objects, arrays, and various data types without issues.
+- **Payload:** large json file.
+- **HTTP Method:** `POST` and `GET`
+- **Deserialization:** Ensure accurate conversion of JSON payloads, that it's performed using the configured JSON libraries (quarkus-rest-jackson, quarkus-rest-jsonb).
+- **HTTP Status Code:** Expect `201 Created` upon successful payload processing.
+- **Response Content:** Validate that the response contains the expected message.
+
+####  Handling invalid JSON payloads:
+Test how the application responds to malformed or invalid JSON requests
+ensuring appropriate error responses are returned.
+- **Payload:** json file without commas between objects, extra strings, etc.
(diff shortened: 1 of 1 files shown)


QQE-379.md
```diff
@@ -0,0 +1,44 @@
...
+### What to test
+
+ - Successful JSON payload processing: Verify that valid JSON payloads requests to REST endpoints and assert that the responses contain the correct JSON data and status codes.
+ - Check Big JSON payload processing: Verify that the application can efficiently process large JSON payloads without issues.
+ - Handling invalid JSON payloads: Test how the application responds to malformed or invalid JSON requests,
+ensuring appropriate error responses are returned.
+ - Handling missing or incorrect data types: Validate the behavior when required JSON fields are missing or have incorrect data types,
+ensuring proper error handling.
+ - Testing complex JSON structures: Evaluate the handling of JSON payloads containing nested objects, arrays, and various data types,
+ensuring serialization and deserialization.
```
mjurc, line 22: I think all those are correct, but maybe it would be nice to have a plan of what should be checked also. In general our tests are rather poor on negative scenarios :)

QQE-379.md
```diff
@@ -0,0 +1,44 @@
...
+
+### What not to test
+
+- Specific JSON Library Configurations: This test plan focuses on the functionality of JSON handling in Quarkus REST endpoints and clients,
+not the specifics of underlying JSON libraries (e.g., Jackson, JSON-B). While both libraries will be used during testing, the aim is to verify the behavior of the Quarkus.
```
mjurc, line 27: Are both of them supported? If they're both supported and recommended, imo we should take them as valid part of testing 'matrix'

If Quarkus is heavily opinionated here or doesn't allow for configuration, I think my point is moot but if they're both supported and configurable, we should be testing it

QQE-379.md
jcarranzan: ok , I'll describe the checks and add negative scenario(s), thanks.

QQE-379.md
jcarranzan: Yes, currently both are supported, although seems from docs that they use Jackson as a first option. And also, I checked this "old" thread discussion--> https://groups.google.com/g/quarkus-dev/c/RTnvH6LSdxs   where they lean towards Jackson. 

Anyways, I could include both as you mentioned.

QQE-379.md
mjurc: Hey, has this been resolved? I do not see much change in the PR right now 😅

QQE-379.md
mjurc: Was there any conclusion on this? I'm gonna unresolve, but maybe I am misunderstanding something.

QQE-379.md
jcarranzan: Hey, I am working on this right now, I'll let u know, thanks.

jcarranzan: I've tweaked the sections with some detailed description and I've added what should be checked also.

QQE-379.md
jcarranzan: Well, I was investigating a bit, and my idea was introduce the TD into the http/http-advance-reactive module and there already is used `quarkus-rest-jsonb` so it's used for serialization support for Quarkus REST.
On the other hand, Jackson serialization quarkus-rest-jackson, it's already used in other tests as in http/jakarta-rest-reactive with advanced features such as `JsonView support` or `@CustomSerialization`.
So from my humble opinion, I would put out of this testing scope .
However, I am open to incorporating alternative approaches or additional tests if you have any suggestions about it.

jcarranzan: I was thinking also, if I could execute same tests using jackson library, we will have the coverage for both libraries, but not sure if I can get it exactly.

jedla97:
As I go review the code in TS i read the TP as well. Could you add or at least test how these tests impact the time execution? As you setting [RestService](https://github.com/quarkus-qe/quarkus-test-suite/pull/1890/files#diff-1831738ab42519d644006bea94012b00ea9c1b489cfb059c4b6ae2b38116adb4R33) with only two classes so it will be probably need to be compiled to native independently.

QQE-379.md
mjurc: Thank you.

mjurc:
Agree with @jedla97 here

jcarranzan:
Ok , I've updated the TD to try test both libraries, and I'll keep with @jedla97 suggestion, thanks.

mjurc (approved):
LGTM, thanks for the fixes

PR #186, opened 2024-09-17
Add TP for QUARKUS-4221 Productization of quarkus-picocli extension

jcarranzan:
### Links

JIRA: https://issues.redhat.com/browse/QUARKUS-4221

PR: https://github.com/quarkus-qe/quarkus-test-suite/pull/2003

Quarkus documentation:

   - Logging
   - Tracing
   - Metrics
   - Security
   - OpenAPI
   - Data sources
   - Frontend
   - Qute

### DIFF
--- a/QUARKUS-4221.md
+++ b/QUARKUS-4221.md
@@ -1,0 +1,103 @@
+# QUARKUS-4221 Productization of Picocli extension
+
+JIRA: https://issues.redhat.com/browse/QUARKUS-4221
+
+PR: https://github.com/quarkus-qe/quarkus-test-suite/pull/2003
+## Scope of the testing
+### What to test
+
+The test coverage will be verified for environments: JVM for Baremetal.
+
+ It won't be tested on Native because this support was requested by the Keycloak team, and they explicitly stated (in the comments of the Jira task) that they don't work with native executables,
+so it's not needed at the moment.
+
+Openshift excluded due to the impracticality of running command-line applications with CLI arguments in these environment.
+
+### Scenarios
+* Verify command line application with some annotations such as: `@CommandLine.Command` / `@CommandLine.Option` 
+
+  - **Purpose:** Ensure that the greeting command works correctly and produces the expected output when provided with the correct arguments.
+  - **Configuration:** Set up a @QuarkusApplication RestService to run the application in a controlled environment.
+  - **Test:** Run the command and verify that the output matches the expected greeting message.
+  
+
+* Verify the usage of proxy scope with @ApplicationScope in beans to check 'Picocli will not be able to set field values in such beans' according 
+the warning in the guide: [simple-command-line-application](https://quarkus.io/guides/picocli#simple-command-line-application).
+
+  - **Purpose:** To test if Picocli can handle beans annotated with `@ApplicationScoped`. 
+  According to Picocli documentation, `@ApplicationScoped` beans should not be used with Picocli commands
+  because Picocli cannot set field values in such beans due to their proxy nature. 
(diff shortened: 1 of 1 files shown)


QUARKUS-4221.md
```diff
@@ -0,0 +1,75 @@
+# QUARKUS-4221 Productization of Picocli extension
+
+JIRA: https://issues.redhat.com/browse/QUARKUS-4221
+
+PR: https://github.com/quarkus-qe/quarkus-test-suite/pull/2003
+## Scope of the testing
+### What to test
+
+The test coverage will be verified for environments: JVM for Baremetal and Openshift.
```
fedinskiy, line 9: Why Openshift? Will you verify any K8s specific behaviors?

fedinskiy, line 9: They mention in the guide, that there are some potential problems on Native. Shouldn't we at least minimally check it?

QUARKUS-4221.md
```diff
@@ -0,0 +1,75 @@
...
+
+* Verify simple command line application with some annotations such as: `@CommandLine.Command` / `@CommandLine.Option`
```
fedinskiy, line 11: Is there a coverage in upstream?

QUARKUS-4221.md
jcarranzan: No, there is nothing specific to verify about K8s, but I thought all our tests should also be executed in Openshift as a part of our testing strategy.

jcarranzan: I omitted on Native because this support was asked by keycloak and in a previous analysis they said don't worry about native (they don't play with Native executables).

QUARKUS-4221.md
jcarranzan: Upstream unit tests coverage is here:
https://github.com/quarkusio/quarkus/tree/main/integration-tests/picocli

QUARKUS-4221.md
fedinskiy: Please mention this reasoning in the plan

fedinskiy: 1. The guide mentions some Openshift/Kubernetes specific features. Maybe we should cover them, especially if there is no coverage in upstream.

jcarranzan: I've added Openshit verification in the scope of the testing, thanks.

QUARKUS-4221.md
fedinskiy: What scenarios a covered here? Is there some, which are already done upstream and so we do not need to add it in our TS?

jcarranzan: Yes, there are some basic tests upstream related to commands that I also use here, but in a slightly different way. In my tests, I use a @QuarkusApplication with a RestService running, so they are tested from a different angle. However, if you believe those upstream tests cover the scenarios well enough and my tests are redundant, I'll remove them.

fedinskiy: @QuarkusApplication and RestService are only used by Quarkus QE, so they aren't of interest to anyone else :-)

It is a task for the author of TP do look into the existing coverage and decide, whether it is enough or not, and if it isn't — add the lacking scenarios. Reviewers task is to check, whether author do not forget about it. Ideally, the current coverage should be described, so anyone who works on this in the future will be able to find our reasoning.

jcarranzan: IMHO these tests extend beyond upstream unit tests by utilizing @QuarkusApplication and @RestService (as a bunch of ours test in several modules), this allows us to perform some kind of integration testing, so I'll keep them and I've described it in the TP.

fedinskiy (approved):

PR #189, opened 2024-10-01
3.15 aarch64 test plan updates

mjurc:
* Updated QUARKUS-3007 to be more in line with what happened in 3.8
* Created followup test plan for OCP on aarch leftovers after 3.8 in QUARKUS-3446 test plan
* Created followup test plan for bare metal coverage in 3.15 in QUARKUS-4942 test plan

### Links

JIRA:
* https://issues.redhat.com/browse/QUARKUS-3007
* https://issues.redhat.com/browse/QUARKUS-3446
* https://issues.redhat.com/browse/QUARKUS-4942

Quarkus documentation: N/A

   - Logging
   - Tracing
   - Metrics
   - Security
   - OpenAPI
   - Data sources
   - Frontend
   - Qute

### DIFF
--- a/QUARKUS-3007.md
+++ b/QUARKUS-3007.md
@@ -7,34 +7,29 @@ for x86-64 architecture.
 
 For the purposes of planning, the following tasks therefore need to be executed (ordered by priority):
 * Same test coverage and certification matrix on OCP on aarch64 for both JVM and native mode as on OCP on x86-64 (see
-  https://issues.redhat.com/browse/QUARKUS-3446)
+  [QUARKUS-3446](QUARKUS-3446.md))
 * Certification of RHBQ native mode on OCP on aarch64
-* Certification of RHBQ in both JVM and native mode in the bare metal scenarios on aarch64
+* Certification of RHBQ in both JVM and native mode in the bare metal scenarios on aarch64 
+  (see [QUARKUS-4942](QUARKUS-4942.md))
 
 ## Scope of the testing
 
 ### General
-* Catch-up on the test coverage on OCP to be same for aarch64 as it is for x86-64
-  * Enabling Serverless tests in `openshift-arm` profile once RH Serverless supports aarch64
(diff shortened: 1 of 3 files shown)


QUARKUS-3007.md
```diff
@@ -65,34 +58,31 @@ execution. We should consider moving the jobs to the same pipeline, managing its
...
-  * For machine time, parallelism in execution should be established. The spin up of beaker machines does not take that
-    long in trials, but we don't have any significantly large sample size yet to say how it will look in long term.
+  * Beaker and OCP provisioning run in parallel and complete within 2 hours.
+  * Test jobs execute in parallel with x86-64 coverage on different machine quota.
+  * Running in parallel with x86-64 coverage, the pipeline should finish within 10 hours.
   * For people/investigation time, this depends on how different JDK and test services act on aarch64. So far, 
     significant functional bugs were not present on aarch64. All the issues historically were related to either missing
-    test services or missing productized dependencies. For native applications, we have no baseline to base our
-    estimates upon.
+    test services or missing productized dependencies.
```
rsvoboda, line 82: Can you pls add overall time impact? So people do not need to sum the times mentioned above.

To me the implication is that the execution takes basically one work day. Investigation may trigger additional day or more, depending on the distribution of the failures and their root cause.

mjurc: I think I've added the execution estimation here - https://github.com/quarkus-qe/quarkus-test-plans/pull/189/files#diff-a60880627ee368f2d459af8d31ad2032f9e95d54958b619d3b73887a746efbdfR78. Basically it doesn't stretch the execution of extended platforms testing pipeline, since it runs in parallel on different infra, but it can run for a while.

Actually, I will remove the line about investigation. This here is the same as for regular bare-metal/OCP testing. There's nothing different here, and I don't think there's a useful estimation to be made here independent on the builds.

rsvoboda (approved):

PR #191, opened 2025-01-17
Add test plan for QQE-1290

mocenas:
### Links

JIRA: https://issues.redhat.com/browse/QQE-1290

Quarkus documentation:

   - Logging
   - Tracing
   - Metrics
   - Security
   - OpenAPI
   - Data sources
   - Frontend
   - Qute

### DIFF
--- a/QQE-1290.md
+++ b/QQE-1290.md
@@ -1,0 +1,19 @@
+# QQE-1290 Test RestClientReactive base URL override
+
+JIRA link: https://issues.redhat.com/browse/QQE-1290
+
+Test plan covers reactive rest client's ability to override base URL.
+
+## Scope of the testing
+Verify that quarkus-reactive-rest-client is able to override base URL by annotation-based parameters on runtime.
+Tests will be executed on baremetal JVM and native.
+
+## Automated test development
+- New tests will be part of already existing test module `rest-client-reactive`
+
+## Impact on resources
+- Baremetal JVM:  ~ 30 seconds.
+- Baremetal native: ~ 1:30 min.
+
+# Links
+Original PR: https://github.com/quarkusio/quarkus/pull/43331


QQE-1290.md
```diff
@@ -0,0 +1,19 @@
+# QQE-1290 Test RestClientReactive base URL override
+
+JIRA link: https://issues.redhat.com/browse/QQE-1290
+
+Test plan covers reactive rest client's ability to override base URL.
+
+## Scope of the testing
+Verify that quarkus-reactive-rest-client is able to override base URL by annotation-based parameters on runtime.
+Tests will be executed on baremetal JVM and native.
```
fedinskiy, line 9: Why not Openshift?

mocenas: Technically we probably can run this on OCP but IMHO makes no sense, since only thing the tests are actually testing is overriding some config inside rest-client - which is independent of runtime env. So IMHO running this on OCP is waste of execution time.

fedinskiy (approved):

QQE-1290.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Automated test development
+- New tests will be part of already existing test module `rest-client-reactive`
+
+## Impact on resources
+- Baremetal JVM:  ~ 30 seconds.
+- Baremetal native: ~ 1:30 min.
```
michalvavrik, line 16: If it is incorporated to an existing test class, then impact should be roughly same for JVM and baremetal, if it is not a new test class, then if executable is reused, impact will be lesser. If custom executable is needed, build will only take 1:30?

I don't care about exact times, it is impossible to measure this properly. But I am interested behind logic of these times, it feels bit random. You measured it?

QQE-1290.md
```diff
@@ -0,0 +1,19 @@
+# QQE-1290 Test RestClientReactive base URL override
+
+JIRA link: https://issues.redhat.com/browse/QQE-1290
+
+Test plan covers reactive rest client's ability to override base URL.
+
+## Scope of the testing
+Verify that quarkus-reactive-rest-client is able to override base URL by annotation-based parameters on runtime.
```
michalvavrik, line 8: `quarkus-reactive-rest-client` does not exist in main or in 3.15. In 3.8 there was `quarkus-rest-client-reactive`, but the name was different to what you are referring to.

QQE-1290.md
mocenas: It is creating new test class. I measured this time on my PC. It is possible that GH/jenkins executors will take more time.

michalvavrik: oki, I must ask for a better machine now that I am getting a new laptop. never had such a build but np

PR #196, opened 2025-02-19
Add QUARKUS-5623 test plan

mocenas:
Add a test plan for QUARKUS-5623

### Links

JIRA: https://issues.redhat.com/browse/QUARKUS-5623

Quarkus documentation: https://quarkus.io/version/main/guides/opentelemetry-logging

Extension/feature community status: dev-preview

 - [X] Make sure you have considered the following areas when preparing the test plan:

   - Logging
   - Tracing
   - Metrics
   - Security
   - OpenAPI
   - Data sources
   - Frontend
   - Qute

### DIFF
--- a/QUARKUS-5623.md
+++ b/QUARKUS-5623.md
@@ -1,0 +1,47 @@
+# QUARKUS-5623 - Logging in OpenTelemetry
+
+JIRA: https://issues.redhat.com/browse/QUARKUS-5623
+
+Upstream PRs:
+- https://github.com/quarkusio/quarkus/pull/38239
+- https://github.com/quarkusio/quarkus/pull/46354
+
+Documentation: https://quarkus.io/version/main/guides/opentelemetry-logging
+
+## Preview status
+OpenTelemetry logging is currently considered dev-preview in quarkus.
+
+## Scope of testing
+- Test OpenTelemetry quarkus extension.
+- Verify that quarkus application is sending its logs to the logging backend.
+  - Example backend uses grafana/otel-lgtm, which is all-in-one solution using grafana for frontend and Loki for log collecting. 
+- Verify that log messages as well as log levels are correct.
+  - Log formatting is not tested as log lines store individual fields for log level, message, time, etc.
+- Log level sending config 
+  - Log lines that are not displayed on other log channels (filtered due to too detailed log level), are not send to the backend as well. 
+  Verify that sending to backend responds to changing log level config. 
+- Bulk log sending - verify bulk size & time of sending. 
+- Verify that hostname (part of log data send) is not fixed at build time, verify for both JVM and native.
+- Verify that custom attributes can be sent using `quarkus.otel.resource.attributes`.
+
+## Getting familiar with the feature
+Following actions were taken to ensure familiarity:
+- Focus on exploratory testing of the feature
+- Ensure good user experience and simplicity of use
+- Check the community status of the extension/feature, preview and experimental ones should be considered carefully
+
+## Automated test development
+- Test suite - New scenarios for baremetal and OCP execution.
+- Test framework - requires implementation of new test service for grafana.
+
+### Impact on TS and resources
+- Two additional test classes in one module - for baremetal and OCP.
+Native build will be only for one test on baremetal and one on OCP.
+Expecting additional time to execute:
+  - OCP JVM - 10 min
+  - OCP native - 10 min
+  - Baremetal JVM - 8 min
+  - Baremetal native - 5 min
+
+## Contacts
+* Tester: Martin Očenáš <mocenas@redhat.com>


QUARKUS-5623.md
```diff
@@ -0,0 +1,36 @@
+# QUARKUS-5623 - Logging in OpenTelemetry
+
+JIRA: https://issues.redhat.com/browse/QUARKUS-5623
+
+Upstream PRs:
+    - https://github.com/quarkusio/quarkus/pull/38239
```
rsvoboda, line 6: Fix the formatting, see for example https://github.com/quarkus-qe/quarkus-test-plans/blob/0ecb4d376d284867df7db238b75df4e5b62c2dc2/QUARKUS-5623.md

QUARKUS-5623.md
```diff
@@ -0,0 +1,36 @@
...
+
+Documentation: https://quarkus.io/version/main/guides/opentelemetry-logging
+
+## Preview status
+OpenTelemetry logging is currently considered dev-preview in quarkus.
+
+## Scope of testing
+Test OpenTelemetry quarkus extension.
+Verify that quarkus application is sending its logs to the Grafana backend.
+Verify that log messages as well as log levels are correct.
```
rsvoboda, line 17: Bullet-points would be better for this part

QUARKUS-5623.md
```diff
@@ -0,0 +1,36 @@
...
+    - https://github.com/quarkusio/quarkus/pull/46354
+
+Documentation: https://quarkus.io/version/main/guides/opentelemetry-logging
+
+## Preview status
+OpenTelemetry logging is currently considered dev-preview in quarkus.
+
+## Scope of testing
+Test OpenTelemetry quarkus extension.
+Verify that quarkus application is sending its logs to the Grafana backend.
```
rsvoboda, line 16: `Grafana backend` ... what exactly do you mean by that. To me Grafana is solution for dashboards, it takes data from a "data source" - https://grafana.com/docs/grafana/latest/datasources/#built-in-core-data-sources

QUARKUS-5623.md
```diff
@@ -0,0 +1,36 @@
...
+Verify that log messages as well as log levels are correct.
+
+## Getting familiar with the feature
+Following actions were taken to ensure familiarity:
+- Focus on exploratory testing of the feature
+- Ensure good user experience and simplicity of use
+- Check the community status of the extension/feature, preview and experimental ones should be considered carefully
+
+## Automated test development
+- Test suite - New scenarios for baremetal and OCP execution.
```
rsvoboda, line 26: `OCP execution` ... so the TF Grafana service will support both baremetal and OCP environments?

QUARKUS-5623.md
```diff
@@ -0,0 +1,36 @@
...
+- Test framework - requires implementation of new test service for grafana.
+
+### Impact on TS and resources
+- Two additional test classes in one module - for baremetal and OCP. Expecting additional time to execute:
+  - OCP - 2 min 30 seconds
```
rsvoboda, line 31: OCP will be JVM only?

QUARKUS-5623.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Getting familiar with the feature
+Following actions were taken to ensure familiarity:
+- Focus on exploratory testing of the feature
+- Ensure good user experience and simplicity of use
+- Check the community status of the extension/feature, preview and experimental ones should be considered carefully
+
+## Automated test development
+- Test suite - New scenarios for baremetal and OCP execution.
+- Test framework - requires implementation of new test service for grafana.
```
rsvoboda, line 27: What about customization of the logs format, is that possible? I think this should be checked.
Also support for sending logs in bulks + configuration of the bulk size / frequency of the sends in case of log heavy application.
What if the service is not available? Or the service changes name/address?

rsvoboda (changes requested):

QUARKUS-5623.md
mocenas: OK, I will clarify this more in the TP. I intend to use grafana/otel-lgtm image, which is basically kind of all-in-one for grafana (display), Loki (logging itself) and more.

QUARKUS-5623.md
mocenas: Yes, I'm already working on this in the FW and it works fine in both envs - https://github.com/quarkus-qe/quarkus-test-framework/pull/1503

QUARKUS-5623.md
mocenas: Formatting is not supported. Getting the message from backend actually gets JSON which has separate fields for severity, timestamp, log message, etc. for each log line. So no formatting applies here.

If service is not available, the exporter will write an error to the console (no exception is thrown), that network connection failed. It will then repeatedly try to reconnect. AFAIK this is nothing specific to the logging module of OTEL. IMHO, since we're not testing this elsewhere for OTEL, and this is nothing extra, it is not necessary to have automated tests for.

I will check the bulk configuration, AFAIK there is no specific configuration for otel-logging, but maybe some generic configuration will apply here.

QUARKUS-5623.md
mocenas: You've linked this very file as an example :-), but I got it and fixed it.

QUARKUS-5623.md
mocenas: Changed.

QUARKUS-5623.md
mocenas: I've added OCP native the list. I works on OCP native so no point in excluding it.

QUARKUS-5623.md
mocenas: I've added description to the TP, I hope it is sufficient.

QUARKUS-5623.md
rsvoboda: `JSON which has separate fields` ... can be the fields configured then? Because I can imagine companies having their own requirements for logging data. We already had several issues for enhancing the "stout" logging customization.
If that's not available, I think it is a good candidate for enhancement GitHub issue

rsvoboda: a digest of things discussed in this thread should be added to the test-plan

mocenas: I don't think these field are configurable, and also I'm not sure if these are managed by OTEL, quarkus or some other logging service (Loki - backend one or maybe something on app side?).

We can create a RFE to have this configurable, but we should probably present some strong reason why.

I'm pasting here an example of json for log line, IMHO  it should be sufficient for most cases
```
"stream": {
          "code_lineno": "-1",
          "detected_level": "info",
          "host_name": "mocenas-pc",
          "log_logger_namespace": "org.jboss.logging.Logger",
          "observed_timestamp": "1740051676317910406",
          "scope_name": "io.quarkus.opentelemetry",
          "service_name": "logging-app",
          "service_version": "1.0.0-SNAPSHOT",
          "severity_number": "9",
          "severity_text": "INFO",
          "telemetry_sdk_language": "java",
          "telemetry_sdk_name": "opentelemetry",
          "telemetry_sdk_version": "1.44.1",
          "thread_id": "1",
          "thread_name": "main",
          "webengine_name": "Quarkus",
          "webengine_version": "999-SNAPSHOT"
        },
        "values": [
          [
            "1740051676334889181",
            "Starting new Quarkus gRPC server (using Vert.x transport)..."
          ]
        ]
```

rsvoboda: +1 to create https://github.com/quarkusio/quarkus/issues/new?template=feature_request.yml

And tag Loic, Roberto and Bruno

rsvoboda:
Please don't forget about `a digest of things discussed in this thread should be added to the test-plan`, I have a few meetings now, so I will check this around 3 PM.

QUARKUS-5623.md
mocenas: RFE created - https://github.com/quarkusio/quarkus/issues/46445

mocenas:
> Please don't forget about `a digest of things discussed in this thread should be added to the test-plan`, I have a few meetings now, so I will check this around 3 PM.

I've added the 'digest' to the scope of testing.

mocenas:
I've added some test cases to scope. Also I've changed this to draft since RFE https://github.com/quarkusio/quarkus/issues/46445 is currently being discussed.

Also new test cases would require changes of properties, meaning app rebuilds (additional builds) which would increase the execution time. E.g. native currently takes about 5 minutes local and 10 minutes on OCP. If we add like 3 builds due to different config, this time can multiply. Is it worth it all this execution time @rsvoboda? Meaning should we go with full JVM/native baremetal/OCP matrix here?

rsvoboda:
> Meaning should we go with full JVM/native baremetal/OCP matrix here?

What would be your proposal to get a balance @mocenas ?

mocenas:
> > Meaning should we go with full JVM/native baremetal/OCP matrix here?
> 
> What would be your proposal to get a balance @mocenas ?

What abound one test for native (just to verify it works) and others just for JVM?

rsvoboda:
Sounds reasonable as this is targeting TP level

mocenas:
> Sounds reasonable as this is targeting TP level

OK, I altered it in the test plan.

rsvoboda:
@mocenas will you drop draft status?

mocenas:
> @mocenas will you drop draft status?

I left draft status because of RFE, how will that go about the log fields. (RFE is still mentioned in the TP). I would like to resolve this somehow before dropping draft status.

rsvoboda:
So talk to the developer, give a day. If things are not concluded or clear, we can proceed with current plan and update the TP later.

rsvoboda (approved):

rsvoboda:
I've approved, can be merged once @fedinskiy rechecks and approves this PR

fedinskiy (approved):

PR #201, opened 2025-03-19
Add QQE-1450 test plan

mocenas:
### Links

Add test plan for websocket next. There is no quarkus JIRA for the issue yet. If one is created we can rename this TP.

Quarkus documentation: 
https://quarkus.io/version/main/guides/websockets-next-reference
https://quarkus.io/guides/websockets-next-tutorial

Extension/feature community status: stable

   - Logging
   - Tracing
   - Metrics
   - Security
   - OpenAPI
   - Data sources
   - Frontend
   - Qute

### DIFF
--- a/QQE-1450.md
+++ b/QQE-1450.md
@@ -1,0 +1,61 @@
+# QQE-1450 Websocket next
+
+## Scope of testing
+Verify quarkus' websocket-next extension.
+Covers both websocket server and client.
+
+### Test cases
+- Smoke test - verify test client can connect to quarkus WS server and exchange messages.
+  - Cover both text and binary messages.
+- Multiple clients exchanging messages though server.
+  - Verify broadcasting messages works for more than 2 clients.
+- Exchange messages using reactive API.
+- Cover telemetry gathering.
+  - OpenTelemetry traces.
+  - Micrometer metrics.
+- Error handling - onError endpoint should be called if:
+  - Runtime exception is thrown on endpoint
+  - Uni/Multi receives a failure
+- Test (de)serialization of object - with both default and custom codec.
+- Ping, pong messages
+  - Verify configuration for periodic ping sending - for both client and server.
+  - Verify that pong messages are send and contain same data as ping.
+- Concurrent processing mode - in this mode, quarkus does not guarantee ordering of events.
+  - Test that long processing of one event, does not block the following one.
+- Subwebsockets - verify these work and have correct HTTP path.
+- Storing custom user data into sessions on server and client APIs.
+- Inspecting HTTP upgrade to WS.
+  - Verify the inspection event is called and upgrade can be rejected.
+- TLS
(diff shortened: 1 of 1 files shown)


rsvoboda:
Replacing myself with Michal as reviewer

fedinskiy (commented):
What about additional topics, eg metrics or security?

Also, I do not really understand what does second sentence in `scope` section should mean.

mocenas:
> What about additional topics, eg metrics or security?

I'm not sure what security tests would add to this. Security at WS basically takes all settings from normal quarkus HTTP server (either you want TLS or authentication), and adds them without much config. I can add these tests, but don't see much value in them.

Quarkus WS supports OTEL traces and micrometer metric. That is probably worth testing. I will add it.

> Also, I do not really understand what does second sentence in `scope` section should mean.

I agree that one is redundant, I will remove it.

QQE-1450.md
```diff
@@ -0,0 +1,38 @@
+# QQE-1450 Websocket next
+
+## Scope of testing
+Verify quarkus' websocket-next extension.
+Covers both websocket server and client.
+
+### Test cases
```
mjurc, line 7: This section feels really bare bone to me right now. Is the intent just to test ping pong type of apps? What about serialization/rest of features mentioned in docs, is there any plan for that? 

I see there's ten types of endpoints, six server APIs and three client APIs just in the docs. What's the plan for those?

Any thoughts about security, maybe even TLS? That's probably closer to real life usage :)

mjurc (changes requested):

rsvoboda:
Just fyi, to see list of websockets (next) related changes in the codebase look into https://github.com/quarkusio/quarkus/pulls?q=is%3Apr+is%3Aclosed+label%3Aarea%2Fwebsockets+-label%3Atriage%2Finvalid+base%3Amain

QQE-1450.md
mocenas: > This section feels really bare bone to me right now. Is the intent just to test ping pong type of apps? What about serialization/rest of features mentioned in docs, is there any plan for that?

I originally wanted to go with "focus on hight level scenarios" approach. But if we want to be exhaustive, I have no problem with that. More scenarios added.

> I see there's ten types of endpoints, six server APIs and three client APIs just in the docs. What's the plan for those?

Honestly it kindof feels to me like the docs likes to put a numbered headline above every small thing. One test scenario covers like 8 of these sections. But I've added more scenarios, it should cover all of these sections.


> Any thoughts about security, maybe even TLS? That's probably closer to real life usage :)

I've added scenarios for authentication, authorization and TLS. (BTW: IMHO in this context "security" is really broad and non-specific term).

mjurc (approved):
Thanks for the updates.

Let's go with this now :)

fedinskiy (approved):

PR #203, opened 2025-03-21
Add QUARKUS-4573 test plan

michalvavrik:
### Links

JIRA: https://issues.redhat.com/browse/QUARKUS-5186

Quarkus documentation: infra stuff

Extension/feature community status: infra stuff

   - Logging
   - Tracing
   - Metrics
   - Security
   - OpenAPI
   - Data sources
   - Frontend
   - Qute

### DIFF
--- a/QUARKUS-5186.md
+++ b/QUARKUS-5186.md
@@ -1,0 +1,51 @@
+# QUARKUS-4573 - Enable Quarkus QE baremetal test execution of QuarkusIO/Quarkus project test coverage against RHBQ builds
+
+JIRA: https://issues.redhat.com/browse/QUARKUS-5186
+
+TD:
+- internal repo (Red Hat GitLab with Jenkins Jobs mainly)
+- https://github.com/quarkus-qe/quarkus-extracted-tests
+- https://github.com/quarkus-qe/quarkus-test-extractor
+
+[Quarkus main project](https://github.com/quarkusio/quarkus) has automated unit and integration test coverage that should verify functional and other requirements.
+Quarkus QE is tasked with Red Hat Build of Quarkus verification, which from practical POV means we need to have capability to run tests against delivered RHBQ artifacts.
+So far our team has tested the product with our own test coverage, however the Quarkus main project should be more exhaustive, and it has additional advantages, like that it is maintained by engineering teams.
+
+Delivered RHBQ bits should be 99.99 % same, therefore the fact that we only executed our tests were sufficient to detect a production process issues that would cause difference between community and product builds.
+The goal of this task is to extend the test coverage used for RHBQ verification by executing the Quarkus main project tests as well.
+The capability run these tests with RHBQ will avoid test development duplications on the QE side for tests run on bare-metal, cleanup Quarkus QE Test Suite, thus decrease maintenance and test execution time.
+
+We wish to follow with adding a capability to run the Quarkus main project tests in OpenShift and elaborate with engineering team on test requirements for noteworthy features 
+(for example in [Working Groups](https://quarkus.io/working-groups/)) as well as on significant bugs, which would allow further Quarkus QE Test Suite cleanup, because
+the test development will already be fully covered by the Quarkus Main project tests.
+
+## Scope of the testing
+Run bare-metal unit and integration tests created by the Quarkus main project contributors as part of general trigger pipeline used for RHBQ verification.
+We will run unit tests and integration tests in JVM mode and integrations tests in native mode.
+
+## Future considerations
+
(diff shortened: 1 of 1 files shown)


QUARKUS-5186.md
```diff
@@ -0,0 +1,44 @@
...
+## Future considerations
+
+- Run adapted or dedicated integration tests created by the Quarkus main project contributors in OpenShift
+- Run tests on Windows
+- Run tests on Aarch64
+- Run tests with Podman instead of Docker
+
+## Existing test coverage
+So far upstream test coverage is run as part of the Quarkus main project PR CI.
+Test are not run with the RHBQ at all, but we verify selected scenarios with _upstream test coverage_ based on the CI results.
```
rsvoboda, line 28: "Test are not run with the RHBQ at all" please expand this to explain why that is so.

Quarkus upstream and RHBQ are 99.9% identical and that's the main motivation for me.

You should also explain that by doing this effort we expect to do some cleanup in our TS and to lower the maintanance cost by utilizing tests maintained by the whole development team.

QUARKUS-5186.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Impact on test suites and testing automation
+New Jenkins jobs will be added to RBHQ 3.15 and 3.20 general test triggers.
```
rsvoboda, line 31: Make it just 3.20+

3.15 will enter maintenance mode once 3.20 is out

mjurc (commented):
Personally I have no more comments besides Rostislav's, I'll re-review after changes have been added

QUARKUS-5186.md
michalvavrik: > Quarkus upstream and RHBQ are 99.9% identical and that's the main motivation for me.

in theory, you cannot say that production team is working for many weeks or months, deliver something and we just trust that whatever they delivered is what they say they delivered. we need verification because you can't know if there is not a flaw in tools that production team uses etc.  The way that Groovy substitutions for dependency replacement works is also the reason why we need to test everything with RHBQ.

> You should also explain that by doing this effort we expect to do some cleanup in our TS and to lower the maintanance cost by utilizing tests maintained by the whole development team.

I actually expect more work, not clean up. We cannot cleanup framework until we solve the cloud native part of the test coverage. We need strategy on what needs to be tested with OCP because right now, strategy as I understand it is that we test _what makes sense_. Which I personally translate as everything with HTTP traffic, so 99 % of stuff. Until we define strategy, I expect this ticket brings more work for us as we need to run & inspect more test results.

I'll add some text / rephrase it to align bit with what you are saying and then you can provide your POV.

michalvavrik: Should be done.

QUARKUS-5186.md
```diff
@@ -0,0 +1,49 @@
...
+
+TD:
+- internal repo (Red Hat GitLab with Jenkins Jobs mainly)
+- https://github.com/quarkus-qe/quarkus-extracted-tests
+- https://github.com/quarkus-qe/quarkus-test-extractor
+
+[Quarkus main project](https://github.com/quarkusio/quarkus) has automated unit and integration test coverage that should verify functional and other requirements.
+Quarkus QE is tasked with Red Hat Build of Quarkus verification, which from practical POV means we need to have capability to run tests against delivered RHBQ artifacts.
+So far our team has tested the product with our own test coverage, however the Quarkus main project should be more exhaustive, and it has additional advantages, like that it is maintained by engineering teams.
+Delivered RHBQ bits should be 99.99 % same, therefore the fact that we only executed our tests were sufficient to detect a production process issues that would cause difference between community and product builds.
```
rsvoboda, line 13: Add new empty line before this so the rendering looks better

QUARKUS-5186.md
```diff
@@ -0,0 +1,49 @@
...
+The goal of this task is to extend the test coverage used for RHBQ verification by executing the Quarkus main project tests as well.
+The capability run these tests with RHBQ will avoid test development duplications on the QE side for tests run on bare-metal, cleanup Quarkus QE Test Suite, thus decrease maintenance and test execution time.
+We wish to follow with adding a capability to run the Quarkus main project tests in OpenShift and elaborate with engineering team on test requirements for noteworthy features
```
rsvoboda, line 16: Add new empty line before this so the rendering looks better

rsvoboda (approved):
Just 2 minor formatting items, approving.

QUARKUS-5186.md
michalvavrik: done

QUARKUS-5186.md
michalvavrik: done

mjurc (approved):

PR #211, opened 2025-04-09
Add test plan for QUARKUS-5614

jedla97:
### Links

JIRA: https://issues.redhat.com/browse/QUARKUS-5614

Quarkus documentation: https://quarkus.io/guides/security-authorize-web-endpoints-reference

Extension/feature community status: supported

Test coverage: https://github.com/quarkus-qe/quarkus-test-suite/pull/2387

   - Logging
   - Tracing
   - Metrics
   - Security
   - OpenAPI
   - Data sources
   - Frontend
   - Qute

### DIFF
--- a/QUARKUS-5614
+++ b/QUARKUS-5614
@@ -1,0 +1,66 @@
+# QUARKUS-5614 Support `@PermissionsAllowed` defined on meta-annotation
+
+JIRA: https://issues.redhat.com/browse/QUARKUS-5614
+
+Support `@PermissionsAllowed` defined on meta-annotation
+
+Upstream PRs:
+- https://github.com/quarkusio/quarkus/pull/43241
+
+Upsteram issue:
+- https://github.com/quarkusio/quarkus/issues/39663
+
+Documentation:
+- https://quarkus.io/guides/security-authorize-web-endpoints-reference
+
+## Scope of testing
+Verify that `@PermissionsAllowed` works correctly when applied on a meta-annotation.
+These tests will cover various scenarios of meta-annotations defined on both methods and classes.
+Furthermore, the tests will validate the combined usage of directly applied `@PermissionsAllowed` annotations and those defined through meta-annotations on a single method, ensuring they operate together.
+Additionally, the direct application of `@PermissionsAllowed` on methods and classes will be tested, as this is currently not covered by QE.
+
+The tests will verify that `@PermissionsAllowed` works correctly when used with Quarkus REST (`quarkus-rest`).
+
+
+## Automated test development
+
+### Definition of roles and permissions
+
+The tests will use 4 roles (for 4 users) and 4 defined permission.
+Each role will be assigned a unique set of permissions.
+These roles will be defined in property file using Elytron security.
+The need for 4 roles and 4 permission is to validate multiple combination and definition.
+
+### Test cases
+- Both definition on meta-annotation and direct use of `@PermissionsAllowed` will have same cases, which should have same results.
+- These cases are:
+  - Two annotation on one method where each have one permission
+  - Two annotation on one method where one have one permission and second have multiple permission
+  - Two annotation on one method where one have one permission and second have multiple permission and defined inclusive (user role must have all permission)
+  - One annotation with multiple permission
+  - One annotation with multiple permission and defined inclusive
+  - Permission with definition of own permission class
+    - One annotation with multiple permission and custom permission class
+    - One annotation with multiple permission, custom permission class and defined inclusive
+  - Permission defined on class
+    - Method without defined permission will use permission defined on class
+    - Method without defined permission will use permission defined on class with defined inclusive
+    - Method override class permission by own definition
+- Testing direct use of `@PermissionsAllowed` and `@PermissionsAllowed` defined using meta-annotation together:
+  -  Two annotation on one method where:
+  - Direct usage have one permission and meta-annotation have multiple permission
+  - Direct usage have one permission and meta-annotation have multiple permission with defined inclusive
+  - Direct usage have multiple permissions and meta-annotation have one permission
+  - Direct usage have multiple permissions with defined inclusive and meta-annotation have one permission
+  - Permission inherited from definition in interface
+    - Direct usage have one permission and meta-annotation have multiple permission
+    - Direct usage have one permission and meta-annotation have multiple permission with defined inclusive
+
+
+### Impact on TS and resources
+The new module `security/permissions` will be added to QE TS.
+This module execution will take around 30 seconds for JVM mode and 3 minutes for native mode.
+For OpenShift the JVM mode will take around 8 minutes and native mode take 14 minutes.
+
+## Contacts
+- Tester: Jakub Jedlička <jjedlick@redhat.com>


QUARKUS-5614
```diff
@@ -0,0 +1,66 @@
...
+
+Upsteram issue:
+- https://github.com/quarkusio/quarkus/issues/39663
+
+Documentation:
+- https://quarkus.io/guides/security-authorize-web-endpoints-reference
+
+## Scope of testing
+Verify that `@PermissionsAllowed` works correctly when applied on a meta-annotation.
+These tests will cover various scenarios of meta-annotations defined on both methods and classes.
```
michalvavrik, line 18: I don't think meta-annotations can work on classes, have you tried it?

michalvavrik: https://quarkus.io/guides/security-authorize-web-endpoints-reference#permission-meta-annotation says method or class so it is ok, my bad

QUARKUS-5614
```diff
@@ -0,0 +1,66 @@
...
+
+### Test cases
+- Both definition on meta-annotation and direct use of `@PermissionsAllowed` will have same cases, which should have same results.
+- These cases are:
+  - Two annotation on one method where each have one permission
+  - Two annotation on one method where one have one permission and second have multiple permission
+  - Two annotation on one method where one have one permission and second have multiple permission and defined inclusive (user role must have all permission)
+  - One annotation with multiple permission
+  - One annotation with multiple permission and defined inclusive
+  - Permission inherted from definition in interface
```
michalvavrik, line 40: if you read https://quarkus.io/guides/security-authorize-web-endpoints-reference#endpoint-security-annotations-and-jakarta-rest-inheritance than you know that while historically some scenarios work (maybe we could say even many), inheritance from interface is not supported.

QUARKUS-5614
```diff
@@ -0,0 +1,66 @@
...
+    - Two annotation on one method where one have one permission and second have multiple permission
+    - One annotation with multiple permission and defined inclusive
+  - Permission with definition of own permission class
+    - One annotation with multiple permission and custom permission class
+    - One annotation with multiple permission, custom permission class and defined inclusive
+  - Permission defined on class
+    - Method inhert defined permission
```
michalvavrik, line 47: What does `inhert` mean? (same question for the line below)

QUARKUS-5614
```diff
@@ -0,0 +1,66 @@
...
+    - Method inhert defined permission with defined inclusive
+    - Method override class permission by own definition
+- Testing direct use of `@PermissionsAllowed` and `@PermissionsAllowed` defined using meta-annotation together:
+  -  Two annotation on one method where:
+    - Dirrect usage have one permission and meta-annotation have multiple permission
+    - Dirrect usage have one permission and meta-annotation have multiple permission with defined inclusive
+    - Dirrect usage have multiple permissions and meta-annotation have one permission
+    - Dirrect usage have multiple permissions with defined inclusive and meta-annotation have one permission
+  - Permission inherted from definition in interface
```
michalvavrik, line 56: same inheritance note as above

QUARKUS-5614
```diff
@@ -0,0 +1,66 @@
...
+    - Dirrect usage have one permission and meta-annotation have multiple permission
+    - Dirrect usage have one permission and meta-annotation have multiple permission with defined inclusive
+
+
+### Impact on TS and resources
+The new module `security/permissions` will be added to QE TS.
```
michalvavrik, line 62: Sadly impl. little differs between REST stacks (Quarkus REST, RESTEasy, WS Next), I am perfectly fine if you are going to test this only with Quarkus REST, but you should specify what you are going to test here.

QUARKUS-5614
michalvavrik: yeah and there is difference between annotations on endpoints and annotations on CDI beans that are not endpoints, again, I am fine if you are going to only test endpoints

michalvavrik (approved):
I left comments, some needs to addressed, some (like inheritance) you might have known but you only need to rephrase it so that we agree. LGTM anyway

fedinskiy (commented):
1. Is there logging/tracing of failed authorisations? If yes, it would be great to check it. If no, I suggest to create an issue for upstream. Monitoring such events can help with protection against attacks or investigate them.

2. There are numerous typos in the text, please fix them (most IDEs should have  natural languages processing plugins)

QUARKUS-5614
jedla97: Thanks I missed this one part of docs. I removed these usecases

QUARKUS-5614
jedla97: Added it to the scope of testing part

jedla97:
> Is there logging/tracing of failed authorisations? If yes, it would be great to check it. If no, I suggest to create an issue for upstream. Monitoring such events can help with protection against attacks or investigate them.

Will check it

> There are numerous typos in the text, please fix them (most IDEs should have natural languages processing plugins)

Should be fixed now

michalvavrik:
> > Is there logging/tracing of failed authorisations? If yes, it would be great to check it. If no, I suggest to create an issue for upstream. Monitoring such events can help with protection against attacks or investigate them.
> 
> Will check it

Depends on the failure. Quarkus throws forbidden exception. If you need more information you can use security events that contain information: which security check failed, security identity, routing context (so you have path etc.) and the exception raised. There is also https://github.com/quarkusio/quarkus/issues/37483 issue that in the end would improve logging, if it ever gets implemented.

michalvavrik:
> > > Is there logging/tracing of failed authorisations? If yes, it would be great to check it. If no, I suggest to create an issue for upstream. Monitoring such events can help with protection against attacks or investigate them.
> > 
> > 
> > Will check it
> 
> Depends on the failure. Quarkus throws forbidden exception. If you need more information you can use security events that contain information: which security check failed, security identity, routing context (so you have path etc.) and the exception raised. There is also [quarkusio/quarkus#37483](https://github.com/quarkusio/quarkus/issues/37483) issue that in the end would improve logging, if it ever gets implemented. And also, you can define custom exception mapper and catch the exception and log it, some users do that and than handle response themselves, but we do not recommend it with OIDC etc.

jedla97:
> Is there logging/tracing of failed authorisations? If yes, it would be great to check it. If no, I suggest to create an issue for upstream. Monitoring such events can help with protection against attacks or investigate them.

@fedinskiy I looked at + looked at sugestions above. From my POV I'm not exactly sure what to test. For tracing I can use micrometer and check if the request was captured (corect ethod, status and path)

For logging I can use `EventObserver` to capture `AuthorizationFailureEvent` and log them as user would do. And after that check the Quarkus logs

From my POV testing any of these solution seems little off topic as I nom sure what should be tested there.
For example I can capture

`@ObservesAsync AuthorizationFailureEvent` and from event log the security identity (user + permissions) and routing context.

But don't see much reason why this should be tested. Or did you have some different usecase in mind?

fedinskiy (approved):

jcarranzan (commented):
It should be a _.md_ file (extension file missing)

PR #226, opened 2025-04-29
Policy configuration for invalid certificates

fedinskiy:
### Links

JIRA: https://issues.redhat.com/browse/QUARKUS-5664

Quarkus documentation: https://quarkus.io/guides/tls-registry-reference

   - Logging
   - Tracing
   - Metrics
   - Security
   - OpenAPI
   - Data sources
   - Frontend
   - Qute

### DIFF
--- a/QUARKUS-5664.md
+++ b/QUARKUS-5664.md
@@ -1,0 +1,29 @@
+# QUARKUS-5664 TLS - Enable Policy Configuration for Expired or Not Yet Valid Certificates
+
+## Scope of testing
+Verify, that the user can configure actions for expired or not-yet-valid certificates in TLS registry 
+
+### Test cases
+- verify, that all three possible values of `quarkus.tls.trust-store.certificate-expiration-policy` property ("ignore","warn" and "reject") are working as expected
+- verify, that "ignore" is a default one
+- verify, that it works for both expired and not-yet-valid certificates
+- verify, that it works in JVM and Native, create draft tests for Openshift as well (in case https://github.com/quarkus-qe/quarkus-test-framework/issues/1052 will be fixed).
+- verify, that this behavior is consistent between certificate reloads
+- verify, that there may be different TLS configs active in the same app
+
+## Existing test coverage
+No coverage
+
+### Impact on test suite
+New test should be added. I recommend to add it to either `security/https module`, or `http/rest-client-reactive`
+
+## Impact on resources
+Running time is expected to be between 1 (in JVM mode) to 6 (native mode) minutes. Additional RAM and CPU will not be required.
+
+## References:
+- jira issue: https://issues.redhat.com/browse/QUARKUS-5664
+- the PR: https://github.com/quarkusio/quarkus/pull/45117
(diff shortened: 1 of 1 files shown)


QUARKUS-5664.md
```diff
@@ -0,0 +1,27 @@
...
+### Impact on test suite
+New test should be added. I recommend to add it to either `security/https module`, or `http/rest-client-reactive`
+
+## Impact on resources
+Running time is expected to be between 1 (in JVM mode) to 6 (native mode) minutes. Additional RAM and CPU will not be required.
+
+## References:
+- jira issue: https://issues.redhat.com/browse/QUARKUS-5664
+- the PR: https://github.com/quarkusio/quarkus/pull/45117
+- The guide. There is feature-specific guide yet, but I've asked Clement to add some notes about this feature https://quarkus.io/guides/tls-registry-reference
```
jedla97, line 24:
```suggestion
- The guide. There is no feature-specific guide yet, but I've asked Clement to add some notes about this feature https://quarkus.io/guides/tls-registry-reference 
```

jedla97 (approved):
The TP looks good to me. Have just one comment.

QUARKUS-5664.md
```diff
@@ -0,0 +1,27 @@
+# QUARKUS-5664 TLS - Enable Policy Configuration for Expired or Not Yet Valid Certificates
+
+## Scope of testing
+Verify, that the user can configure actions for expired or invalid certificates in TLS registry
```
michalvavrik, line 4: Here and below you use term `invalid` but I think as far as validity goes, only "not yet valid" invalidity is allowed. I think you need to be more specific, certificates can be invalid for other reasons.

michalvavrik, line 4: Here and below you use term `invalid` but I think as far as validity goes, only "not yet valid" invalidity is allowed. I think you need to be more specific, certificates can be invalid for other reasons.

michalvavrik (commented):
The plan is OK, but I am missing any mention of documentation in your plan, I have reviewed upstream PR and docs you linked and I can't see anything documented, only configuration property is added and it has javadoc that is obviously shown in the docs. That javadoc literally says `When enabled, the certificate expiration date is verified and the certificate (or any certificate in the chain) is rejected if it is expired.`, can this leave questions in our customers? Questions like:

- it this policy applied once (is validation done), on startup?
- when I configure `certificate-expiration-policy=reject` for a named TLS configuration used by a REST client using a certificate that is valid on application startup but invalid on the first use of a REST client, is the trust store rejected?
- if the certificate is expired an hour into application running, what happens? Is this feature all about startup or everytime this certificate is used?

Things like that, I am just fabulating and I think I can answer already, but the point is that docs doesn't go into details. I'd like you to verify docs and mention it in the TP.

Additionally, I'd expect you to test:

- certificate reloading and verify that the policy is applied on the refreshed certificates as well (this can happen for example when a REST client is using OpenShift certificate serving)
- multiple named TLS config with different policies configured at once (in a single app)

michalvavrik (approved):

PR #228, opened 2025-04-30
Add TP for QUARKUS-5662

mocenas:
### Links

JIRA: https://issues.redhat.com/browse/QUARKUS-5662

Quarkus documentation:  https://quarkus.io/guides/security-oidc-code-flow-authentication#using-the-authorization-code-flow-mechanism ,  https://quarkus.io/guides/security-openid-connect-client-reference#jwt-bearer

Extension/feature community status: SUPPORTED

   - Logging
   - Tracing
   - Metrics
   - Security
   - OpenAPI
   - Data sources
   - Frontend
   - Qute

### DIFF
--- a/QUARKUS-5662.md
+++ b/QUARKUS-5662.md
@@ -1,0 +1,30 @@
+# QUARKUS-5667: OIDC and OIDC Client: Support JWT bearer client authentication using client assertion loaded from filesystem
+
+Jira: https://issues.redhat.com/browse/QUARKUS-5662
+
+Upstream documentation:
+* https://quarkus.io/guides/security-oidc-code-flow-authentication#oidc-provider-client-authentication
+* https://quarkus.io/guides/security-openid-connect-client-reference#jwt-bearer
+
+Code changes: 
+* https://github.com/quarkusio/quarkus/pull/45131
+
+This feature cover usage of JWT token loaded from file on filesystem.
+There are two cases - using token in general oidc settings and in oidc-client.
+
+## Scope of the testing
+Tests will require OIDC provider (keycloak in our case) to provide and validate JWT tokens. 
+
+### General
+* A new tests will be implemented in `security/keycloak-oidc-client-extended`.
+* Tests will cover these token-cases for both oidc and oidc-client:
+  * Non-existing file with token.
+  * Empty file with token.
+  * File containing other data than token.
+  * File containing valid token.
+
+### Impact on resources
+* New test should be executed within a minute or single minutes for both bare metal JVM and native
+
+## Contacts
+* Tester: Martin Ocenas <mocenas@redhat.com>


mocenas:
I intentionally left out the OCP for now, while it should be the primary goal of this feature. Because I'm not sure if the following workflow is viable with our framework:

1. Deploy keycloak instance to OCP
2. Get a JWT token from keycloak
3. Deploy OCP secret with the JWT token
4. Deploy the app with the secret mounted

Can our FW support this flow (in the given order)  @michalvavrik ?

michalvavrik (approved):
Approving, because there is nothing wrong in the TP, but it is very short :-))) It will give you flexibility later, I appreciate that before you start writing tests, it is hard to make plans on a difficult topic.

michalvavrik:
> Can our FW support this flow (in the given order) @michalvavrik ?

not without adjustments, you would have to write everything yourself; I agree that OCP would be primary target, but I am not sure that Keycloak is. IIRC the tests I wrote were quite a stretch.

Main motivation for this feature was mounting https://kubernetes.io/docs/concepts/security/service-accounts/#authenticating-credentials and later making sure that tokens are refreshed when K8 mounts a new token (I think it happens when token age is at 80 %). And I did try it over one weekend, but I had to break our QE OCP cluster to make it work, because I had to make some cluster wide changes. Sorry, I don't remember details, you will have to invest time into that and probably check it. But I think that is what you should preferably test, not Keycloak.

mocenas:
> Main motivation for this feature was mounting https://kubernetes.io/docs/concepts/security/service-accounts/#authenticating-credentials and later making sure that tokens are refreshed when K8 mounts a new token (I think it happens when token age is at 80 %). And I did try it over one weekend, but I had to break our QE OCP cluster to make it work, because I had to make some cluster wide changes. Sorry, I don't remember details, you will have to invest time into that and probably check it out. But I think that is what you should preferably test, not Keycloak.

I'm not sure I get the bigger picture here. Why to load the serviceaccount's token into the OIDC client of all things? Is it expected scenario that OCP/K8S will server as an identity provider for some app running in the pod? I've seen similar scenario somewhere, but I'm not sure how common it is.

mocenas:
> > Main motivation for this feature was mounting https://kubernetes.io/docs/concepts/security/service-accounts/#authenticating-credentials and later making sure that tokens are refreshed when K8 mounts a new token (I think it happens when token age is at 80 %). And I did try it over one weekend, but I had to break our QE OCP cluster to make it work, because I had to make some cluster wide changes. Sorry, I don't remember details, you will have to invest time into that and probably check it out. But I think that is what you should preferably test, not Keycloak.
> 
> I'm not sure I get the bigger picture here. Why to load the serviceaccount's token into the OIDC client of all things? Is it expected scenario that OCP/K8S will serve as an identity provider for some app running in the pod? I've seen similar scenario somewhere, but I'm not sure how common it is.

michalvavrik:
That was motivation mentioned in https://github.com/quarkusio/quarkus/issues/40013. Yes, this feature is not specific for any OIDC identity provider, so your TP is OK.

michalvavrik (approved):

QUARKUS-5662.md
```diff
@@ -0,0 +1,26 @@
...
+
+This feature cover usage of JWT token loaded from file on filesystem.
+There are two cases - using token in general oidc settings and in oidc-client.
+
+## Scope of the testing
+Tests will require OIDC provider (keycloak in our case) to provide and validate JWT tokens. 
+
+### General
+* A new tests will be implemented in `security/keycloak-oidc-client-extended`.
+* Tests will cover missing, invalid and valid tokens in files, for both oidc and oidc-client.
```
rsvoboda, line 20: Please clarify what exactly does `invalid` mean.

Expired token? Corrupted token?

Add test coverage for invalid `quarkus.oidc.credentials.jwt.token-path` and `quarkus.oidc.credentials.jwt.token-path` pointing to a binary file or/and file not using the expected format.

As an exploratory task please check what happens if you point to a huge file, e.g. ISO image.

rsvoboda (changes requested):

QUARKUS-5662.md
rsvoboda: Also check what error message get printed if the file exists, but user doesn't have read permission.
(As an exploratory check ...)

michalvavrik: > As an exploratory task please check what happens if you point to a huge file, e.g. ISO image.

:-) what token can be a huge file?

rsvoboda: it can be intentional attack, I would expect OOM

michalvavrik: if you have these sort of rights, you can do a lot of things, e.g. you can write such a huge file that it takes all the FS space? feel free to test your suggestion, I just wonder if we test this way other features that use fs, because I'd expect that almost (almost because I don't know) noone in Quarkus protects against that.

rsvoboda: security needs to lead :)

rsvoboda: on more serious note, it's suggested for exploratory tests, not to create automated test coverage for it

michalvavrik: np, let's do it

mocenas: I added and clarified the test cases. I'm not sure what you meant by "invalid path" @rsvoboda as pretty much any sequence of characters in linux is technically a valid path, so I added missing file scenario (path pointing to non-existing file).
I was also considering to test expired tokens, but that would be problematic, since I will keycloak on bare metal, and new run means new keycloak instance, which would AFAIK not accept tokens from previous one. So we cannot pre-generate tokens and we don't want to wait ten minutes in tests for token to expire.

As for exploratory testing, it works as expected
- permission denied file would correctly throw AccessDeniedException
- Huge token file (about 6GB in my case) would throw `java.lang.OutOfMemoryError: Required array size too large`
So no problems here

rsvoboda (approved):

PR #232, opened 2025-05-26
Add TP for QUARKUS-5858

mocenas:
### Links

JIRA: https://issues.redhat.com/browse/QUARKUS-5858

Quarkus documentation: https://quarkus.io/version/main/guides/security-oidc-bearer-token-authentication#demonstrating-proof-of-possession-dpop

Extension/feature community status: tech-preview

   - Logging
   - Tracing
   - Metrics
   - Security
   - OpenAPI
   - Data sources
   - Frontend
   - Qute


QUARKUS-5858.md
```diff
@@ -0,0 +1,30 @@
+# QUARKUS-5858 Tech preview Support for OAuth2 Demonstrating Proof of Possession
+
+JIRA: https://issues.redhat.com/browse/QUARKUS-5858
+
+Upstream PR: https://github.com/quarkusio/quarkus/pull/45891
+
+Upstream documentation:
+* https://quarkus.io/version/main/guides/security-oidc-bearer-token-authentication#demonstrating-proof-of-possession-dpop
+
+## Preview status
```
michalvavrik, line 10: Regarding your statement that DPoP is considered a tech-preview in Quarkus, I don't believe it. Please answer:

- where can users read it in community docs?
- where is it stated in upstream PR https://github.com/quarkusio/quarkus/pull/45891 or in https://github.com/quarkusio/quarkus/issues/42115

I think you just think that from the facts that:

- https://issues.redhat.com/browse/QUARKUS-5858 is titled _Tech preview Support for OAuth2 Demonstrating Proof of Possession_ but we should:
  -  confirm why it was decided to be a TP and track conditions when this stops being TP, I can see in the history that Thomas changed it to TP manually, we should confirm with him that he did it to stay as TP, not just to allow us to postpone processing because before the release we had a lot to do and Thomas was supposed to prioritize tasks and we asked him to mark unprocessed JIRAs as TPs
  - if true, it is only true for RHBQ and not Quarkus
  - this needs to be tracked as TP only, so I guess part of the JIRA verification must be checking RNs section where this is documented that it is in TP section
- Keycloak offers DPoP as TP, but this feature is not Keycloak specific and https://datatracker.ietf.org/doc/html/rfc9449 is published

TL;DR; at the very least you need to rewrite the sentence, as it stands it is not true for Quarkus but maybe for RHBQ.

QUARKUS-5858.md
```diff
@@ -0,0 +1,30 @@
...
+DPoP is currently considered tech-preview in quarkus.
+
+## Scope of testing
+Verify quarkus works with authentication method "Demonstrating Proof of Possession" (DPoP)
+A new test class will be implemented in `security/keycloak-oidc-client-extended`.
+Tests will verify authentication using DPoP, which will cover test cases:
+* Happy path - correctly accessing resource using DPoP authentication.
+* Authentication failures:
+  * No DPoP header.
```
michalvavrik, line 19: Good, but just having unauthorized doesn't feel enough, let's also try sending valid token using `Authorization` scheme while DPoP is required, so that we know that it is always enforced (judging by the code it is...)

QUARKUS-5858.md
```diff
@@ -0,0 +1,30 @@
...
+  * DPoP proof signed with wrong key.
+  * Mismatching HTTP method in DPoP proof.
+  * Mismatching HTTP path in DPoP proof.
+
+### Impact on resources
+A new module will be created, requires additional test execution.
+Tests will be executed on both JVM and native and on both baremetal and OCP.
+Expected additional execution time in minutes for each case.
```
michalvavrik, line 27: I expect minutes per each test class, so if you plan to have case per class, you should mention how many test cases you expect. Otherwise without saying how many classes or cases  you are adding, we don't know how much you add minutes.

mocenas: In this context by "each case" I mean - few minutes for JVM bare metal + few minutes for OCP native .... So in total 4times "few minutes".
I can break it down into all the cases if you want.

QUARKUS-5858.md
mocenas: You're right I got fooled by JIRA + RN. I was not able to find it in upstream. I removed this.

QUARKUS-5858.md
mocenas: I've added this case.

QUARKUS-5858.md
michalvavrik: I am ok with it now that I understand what you meant

QUARKUS-5858.md
```diff
@@ -0,0 +1,28 @@
...
+
+JIRA: https://issues.redhat.com/browse/QUARKUS-5858
+
+Upstream PR: https://github.com/quarkusio/quarkus/pull/45891
+
+Upstream documentation:
+* https://quarkus.io/version/main/guides/security-oidc-bearer-token-authentication#demonstrating-proof-of-possession-dpop
+
+## Scope of testing
+Verify Quarkus works with authentication mechanism "Demonstrating Proof of Possession" (DPoP)
```
michalvavrik, line 11:
```suggestion
Verify Quarkus works with authentication mechanism "Demonstrating Proof of Possession" (DPoP).
```

michalvavrik (approved):

rsvoboda (approved):
LGTM!

PR #245, opened 2025-07-23
Add test plan for QUARKUS-6241

michalvavrik:
### Links

JIRA: https://issues.redhat.com/browse/QUARKUS-6241

Quarkus documentation: https://quarkus.io/guides/hibernate-reactive#orm-and-reactive-extension-simultaneously

Extension/feature community status: Hibernate ORM supported, Hibernate Reactive preview so preview on average

   - Logging
   - Tracing
   - Metrics
   - Security
   - OpenAPI
   - Data sources
   - Frontend
   - Qute

### DIFF
--- a/QUARKUS-6241.MD
+++ b/QUARKUS-6241.MD
@@ -1,0 +1,52 @@
+# QUARKUS-6241 - Allow Hibernate ORM and Hibernate Reactive to be used in the same application
+
+JIRA: https://issues.redhat.com/browse/QUARKUS-6241
+
+Upstream issue: https://github.com/quarkusio/quarkus/issues/13425
+
+Upstream PR: https://github.com/quarkusio/quarkus/pull/44473
+
+Before Quarkus 3.22, it wasn't possible to use the Quarkus Hibernate ORM extension with the Quarkus Hibernate Reactive extension in one project.
+Now, users can leverage the combination of Hibernate Reactive and Hibernate ORM.
+Support for respective Panache versions will come with Panache 2.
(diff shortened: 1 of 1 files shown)


QUARKUS-6241.MD
```diff
@@ -0,0 +1,52 @@
...
+Resources need to be secured so that we can verify that authentication and authorization are working properly.
+
+## Existing test coverage
+
+Upstream test coverage includes unit tests for default persistence units, named persistence units, and their combination.
+There are also integration tests that are executed in the native mode.
+
+### Impact on test suites and testing automation
+
+A new `security/jpa-reactive` module will be added.
```
mocenas, line 28: Should this really be in a security group? Main focus is coexistence of hibernate ORM and hibernate reactive. It makes more sense to me to put it to DB modules.

michalvavrik: I am bit worried to same something because in the last TP I was criticized for adding there my opinions, but I don't agree, sorry. What the upstream PR documents (using HR for reactive endpoints and H ORM for imperative endpoints) is very edge case. We already know that using Hibernate Reactive brings advantage to very small number of applications, it is definitely not true that you should use it every time you write reactive endpoint. So it only makes sense for apps that cares about latency and vertical scaling on some endpoints while supporting traditional endpoints in the same app. How frequent it is?

What I suggest is much more practical, I say "use Hibernate Reactive to prevent thread switching during auth/authz" because 90 % + cases (basically only when users force it or use awkward extension) run on IO thread.

So it is in `security/jpa-reactive` because we already have `security/jpa` there.

PR #303, opened 2026-06-11
AGENTS.md for assisted review

rsvoboda:
AGENTS.md for assisted review

CC @mjurc wdyt

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,165 @@
+# Agent Instructions for Creating Quarkus Test Plans
+
+Use this file as operating guidance when drafting or revising a Quarkus feature test plan in this repository.
+
+Think like a senior engineer evaluating a feature for production use.
+
+Produce a test plan that is:
+- easy to review
+- easy to execute
+- explicit about scope, ownership, risks, and resource impact
+- aligned with repository conventions and strong existing examples
+
+## Repository Overview
+
+This repository contains test plans for Quarkus product features and enhancements. It is a documentation-focused repository
+that tracks quality engineering efforts, test coverage strategies, and testing automation plans for the Quarkus framework.
+
+## Mandatory repository conventions
+
+- Name the file with the issue key, for example `QUARKUS-123.md`
+- Use PR and commit titles in the form `QUARKUS-ABC - TITLE_OF_THE_RFE`
+
+## Required structure
+
+Unless there is a strong reason not to, generate test plans with these sections in this order:
+
+1. Title with issue key and feature name
+2. JIRA link
+3. Quarkus documentation and upstream links
+4. Short feature summary
+5. Scope of testing
+6. Getting familiar with the feature
+7. Existing coverage
+8. Planned coverage
+9. Impact on test suites and automation
+10. Impact on resources
+11. Open questions or future considerations
+12. Contacts
+13. References
+
+## Authoring rules
+
+### 1. Start with a short feature summary
+State:
+- what the feature is
+- why it matters
+- important support limits or constraints
+
+Do not start with vague background text.
+
+### 2. Add all important links
+Include when available:
+- JIRA
+- Quarkus documentation
+- upstream issues
+- upstream PRs
+- follow-up PRs
+- related test plans
+- community status
+
+Do not force reviewers to search for core references.
+
+### 3. Define scope precisely
+State:
+- what is in scope
+- what is out of scope
+- target environments
+- target modes such as JVM, native, dev mode, OpenShift
+
+Do not leave scope implied. Native mode justification can be skipped because Quarkus fully supports both JVM and Native mode.
+
+### 4. Assess extension interactions and integration points explicitly
+Consider and document how the feature interacts with other Quarkus extensions and cross-cutting concerns.
+Think about:
+- Which extensions users might combine with this feature
+- Common integration patterns (observability, security, data access, etc.)
+- Potential conflicts or special considerations
+
+### 5. Separate existing coverage from planned coverage
+Document:
+- what is already tested upstream
+- what is already tested in QE
+- what new coverage will be added
+
+Do not mix current state with planned work.
+
+### 6. Be specific and justify coverage decisions
+Name exact test targets (modules, platforms, architectures) and explain why each was chosen.
+Examples:
+- "Testing module X because it represents typical usage"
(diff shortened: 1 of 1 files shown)


jedla97 (approved):
Reading this I think it could give agent some clear context what should be done

AGENTS.md
```diff
@@ -0,0 +1,130 @@
+# Agent Instructions for Creating Quarkus Test Plans
+
+Use this file as operating guidance when drafting or revising a Quarkus feature test plan in this repository.
+
+Think like a senior engineer evaluating a feature for production use.
+
+Produce a test plan that is:
+- easy to review
+- easy to execute
```
mjurc: I don't think those two are correct. Goal of test plan is:
- to provide a high level overview of a feature
- to consider how to verify the feature completeness and usability in the feature release
- to ensure the feature fulfills product requirements for RH/IBM products
- to create a plan how to ensure there's no regressions in support for all RHBQ/IBQ supported platforms in z stream

Naturally all of those topics are quite complicated (otherwise we wouldn't have users seeking out our framework to use).

But besides the points above, I think the real form requirements for the test plan are being technically precise, without room for ambiguity and readable.

AGENTS.md
```diff
@@ -0,0 +1,130 @@
+# Agent Instructions for Creating Quarkus Test Plans
+
+Use this file as operating guidance when drafting or revising a Quarkus feature test plan in this repository.
+
+Think like a senior engineer evaluating a feature for production use.
```
mjurc: This will just change the vocabulary register. I don't really mind the language we used in test plans previously, but this I guess will result in the usual ticks that LLMs display

AGENTS.md
```diff
@@ -0,0 +1,130 @@
...
+Produce a test plan that is:
+- easy to review
+- easy to execute
+- explicit about scope, ownership, risks, and resource impact
+- aligned with repository conventions and strong existing examples
+
+## Repository Overview
+
+This repository contains test plans for Quarkus product features and enhancements. It is a documentation-focused repository
+that tracks quality engineering efforts, test coverage strategies, and testing automation plans for the Quarkus framework.
```
mjurc:
```suggestion
This repository contains test plans for Quarkus product features and enhancements. The test plan is a document that contains high level overview of the feature, analysis on ensuring that the feature is complete, usable, compliant with general product requirements, and if needed, special considerations for supporting the feature on all the supported platforms and in z-streams.
```

AGENTS.md
```diff
@@ -0,0 +1,130 @@
+# Agent Instructions for Creating Quarkus Test Plans
```
mjurc:
```suggestion
# Agent instructions for creating Quarkus test plans
```

AGENTS.md
```diff
@@ -0,0 +1,130 @@
...
+
+Think like a senior engineer evaluating a feature for production use.
+
+Produce a test plan that is:
+- easy to review
+- easy to execute
+- explicit about scope, ownership, risks, and resource impact
+- aligned with repository conventions and strong existing examples
+
+## Repository Overview
```
mjurc:
```suggestion
## Repository overview
```

AGENTS.md
```diff
@@ -0,0 +1,130 @@
...
+- maintenance burden
+
+Rough estimates are acceptable if clearly explained.
+
+## Review guidelines
+
+When reviewing test plans:
+- Inspect referenced implementation and documentation
+- Review test plan with authoring rules in mind
+- Provide a concise summary (4-5 bullet points) of areas that are covered well in the test plan
```
mjurc, line 133: I'm just afraid that the reviews themselves are going to be longer than vast majority of the test plans :)

AGENTS.md
```diff
@@ -0,0 +1,130 @@
...
+- Provide a concise summary (4-5 bullet points) of areas for improvement, and add a priority prefix to each bullet point
+- Provide a rating (0-10) of the test plan, 10 means excellent
```
mjurc, line 135: LLMs are bad at onthology and at maths. I'd rather avoid having LLM do ratings on reviews, because it's going to end up rating the vocabulary and matching the style but not the contents of what the model is matching with and I don't think it's particularly useful

mjurc (commented):
I am not sure if this is useful in reaching the outcome we wanted to get when we started writing the test plans. 

Writing a good technical text that others can understand and pick up is one of the hardest things we do, and we need to hone that skill in the people. 

Other thing I am concerned about is that we had tendencies in the team to fall back to doing things only by the checklist in analysis (and using that as a defence to any review) and new feature work, and having LLM scoring/styling the test plans I think is going to make it all the worse for reviewing more complicated things.

I think the 'Authoring rules' section is a good general guideline for writing the test plans though :)

AGENTS.md
rsvoboda: I don't see the use of "AI tools" as mandatory. But better to have a defined strategy if the reviewer uses Claude or Bob.

Sounds good or do you have other concerns before resolving this comment @mjurc ?

AGENTS.md
rsvoboda: So do you propose to drop this line completely?

AGENTS.md
rsvoboda: I like the idea of rating tbh. Maybe extending the instruction to focus on plan completeness with respect to the developed feature?

rsvoboda: And to be clear, rating is just a signal, the review can't be driven by just it.

AGENTS.md
rsvoboda: Ok, I will need to think a bit more about how to formulate this. I may just steel your bullet points in the end :)

mjurc: I think the `technically precise, without room for ambiguity` would be better in place of easy to review/execute

AGENTS.md
mjurc: Oh, this was about number of bullet points itself 😅 Sure I agree with your response

AGENTS.md
mjurc: > I like the idea of rating tbh. Maybe extending the instruction to focus on plan completeness with respect to the developed feature?

I guess we can try and see how it works. My concern is we'll get used to signaling from numbers that are not showing what we need

mjurc (approved):
Thanks for the updates, lgtm

PR #304, opened 2026-06-15
QUARKUS-7174 Add support for OAuth 2.0 Pushed Authorization Requests to OIDC extension

jedla97:
### Links

JIRA: https://issues.redhat.com/browse/QUARKUS-7174

Quarkus documentation: https://quarkus.io/guides/security-oidc-code-flow-authentication#pushed-authorization-requests

Extension/feature community status: Stable

   - Logging
   - Tracing
   - Metrics
   - Security
   - OpenAPI
   - Data sources
   - Frontend
   - Qute

### DIFF
--- a/QUARKUS-7174.md
+++ b/QUARKUS-7174.md
@@ -1,0 +1,48 @@
+# QUARKUS-7174 Add support for OAuth 2.0 Pushed Authorization Requests to OIDC extension
+
+JIRA: https://issues.redhat.com/browse/QUARKUS-7174
+
+Add support for OAuth 2.0 Pushed Authorization Requests to OIDC extension
+
+Upstream PRs:
+- https://github.com/quarkusio/quarkus/pull/51799
+
+Documentation:
+- https://quarkus.io/guides/security-oidc-code-flow-authentication#pushed-authorization-requests
+
+## Scope of testing
+The QUARKUS-7174 adding the ability to use Pushed Authorization Requests (PAR).
+The PAR is enhancement for OAuth 2.0 and OpenID Connect.
+The normal workflow of authorization works as they redirect user to login form and values for parameter as `client_id`, `state` and other are exposed in url.
+This can be altered, captured etc.
+The PAR works similar but on the initial request application push the request directly to authorization server and the server return redirect address without the exposed value, there will be only identifier in url.
+
+Testing of PAR will be done using Keycloak as authorization server.
+The tests will include check if the Quarkus correctly redirect url containing only the identifier and not the values.
+Further tests will verify whether the user has access to secure endpoints using `client_secret` and JWT.
+
+This feature also contains the case where `quarkus.oidc.authentication.par.enabled` is not set and PAR is enabled by default, when the authorization server contains `require_pushed_authorization_requests` in discovery metadata.
+This case won't be tested, as it's not possible to set `require_pushed_authorization_requests` to `true` at `/realms/<realn-name>/.well-known/openid-configuration` using Keycloak.
+
+
+## Automated test development
+
+Upstream coverage:
+* There are 3 tests, covering basic functionality of this feature.
+
+The tests will cover these scenarios:
+* Check if Quarkus and Keycloak communicate using back channel and Quarkus redirect user to url with identifier only
+* Check when `quarkus.oidc.authentication.par.enabled` is set, the user is authorized access secured endpoint.
+* Check when `quarkus.oidc.authentication.par.enabled` is set, the unauthorized user cannot access the secure endpoint.
+* Set the `quarkus.oidc.authentication.par.path` to non-existent endpoint to verify, that this property is respected.
+  * The Keycloak don't have option to change `pushed_authorization_request_endpoint` so this property can be tested only for negative results.
+
+
+### Impact on TS and resources
+The tests will be added to [`keycloak-webapp`](https://github.com/quarkus-qe/quarkus-test-suite/tree/main/security/keycloak-webapp) and [`keycloak-jwt`](https://github.com/quarkus-qe/quarkus-test-suite/tree/main/security/keycloak-jwt) modules.
+
+The tests will be executed on bare-metal and OpenShift in JVM and Native mode.
+The expected time increase will be standard as there won't be addition of module and the PAR properties are not build time properties.
+
+## Contacts
+- Tester: Jakub Jedlička <jjedlick@ibm.com>


QUARKUS-7174.md
```diff
@@ -0,0 +1,46 @@
...
+Upstream PRs:
+- https://github.com/quarkusio/quarkus/pull/51799
+
+Documentation:
+- https://quarkus.io/guides/security-oidc-code-flow-authentication#pushed-authorization-requests
+
+## Scope of testing
+The QUARKUS-7174 adding the ability to use Pushed Authorization Requests (PAR).
+The PAR is enhancement for OAuth 2.0 and OpenID Connect.
+The normal workflow of authorization works as they redirect user to login form and values for parameter as `clint_id`, `state` and other are exposed in url.
```
mocenas:
```suggestion
The normal workflow of authorization works as they redirect user to login form and values for parameter as `client_id`, `state` and other are exposed in url.
```

QUARKUS-7174.md
```diff
@@ -0,0 +1,46 @@
...
+## Automated test development
+
+Upstream coverage:
+* There are 3 tests, covering basic functionality of this feature.
+
+The tests will cover these scenarios:
+* Check if Quarkus and Keycloak communicate using back channel and Quarkus redirect user to url with identifier only
+* Check when `quarkus.oidc.authentication.par.enabled` is set, the user is authorized access secured endpoint.
+* Check when `quarkus.oidc.authentication.par.enabled` is set, the unauthorized user cannot access the secure endpoint.
+
```
mocenas, line 39: Maybe add a test case about PAR endpoint discovery? According to the docs you can specify the endpoint via property `quarkus.oidc.authentication.par.path` or it can be discovered from OAuth metadata `pushed_authorization_request_endpoint`. 
Can we test this?

QUARKUS-7174.md
```diff
@@ -0,0 +1,46 @@
...
+The PAR works similar but on the initial request application push the request directly to authorization server and the server return redirect address without the exposed value, there will be only identifier in url.
+
+Testing of PAR will be done using Keycloak as authorization server.
+The tests will include check if the Quarkus correctly redirect url containing only the identifier and not the values.
+Further tests will verify whether or not the user has access to secure endpoints.
+
+This feature also contains the case where `quarkus.oidc.authentication.par.enabled` is not set and PAR is enabled by default, when the authorization server contains `require_pushed_authorization_requests` in discovery metadata.
+This case won't be tested, as it's not possible to set `require_pushed_authorization_requests` to `true` at `/realms/<realn-name>/.well-known/openid-configuration` using Keycloak.
+
+
```
mocenas, line 27: Just our of curiosity. What will happen if PAR fails? I mean scenario where app have it enabled and AS doesn't support it, or app has wrong endpoint. Will the app fallback to another form of authentication? 

Since the parameter is named `par.enabled` not `par.required` I'm not sure what is correct behaviour. Since PAR is supposed to boost security, it would IMHO be bad if it just falled back to less secure authentication method.

I'm not sure if this worth creating an automated test. But at least manually checking it is IMHO worth it.

mocenas (changes requested):
Generally it is IMHO OK. Found one typo and few suggestions.

QUARKUS-7174.md
```diff
@@ -0,0 +1,46 @@
...
+This can be altered, captured etc.
+The PAR works similar but on the initial request application push the request directly to authorization server and the server return redirect address without the exposed value, there will be only identifier in url.
+
+Testing of PAR will be done using Keycloak as authorization server.
+The tests will include check if the Quarkus correctly redirect url containing only the identifier and not the values.
+Further tests will verify whether or not the user has access to secure endpoints.
+
+This feature also contains the case where `quarkus.oidc.authentication.par.enabled` is not set and PAR is enabled by default, when the authorization server contains `require_pushed_authorization_requests` in discovery metadata.
+This case won't be tested, as it's not possible to set `require_pushed_authorization_requests` to `true` at `/realms/<realn-name>/.well-known/openid-configuration` using Keycloak.
+
```
jcarranzan, line 26: The upstream PR use 2 authentication methods (client_secret and JWT),  it would make sense to cover them both as well here in my opinion.

QUARKUS-7174.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Automated test development
+
+Upstream coverage:
+* There are 3 tests, covering basic functionality of this feature.
+
+The tests will cover these scenarios:
+* Check if Quarkus and Keycloak communicate using back channel and Quarkus redirect user to url with identifier only
+* Check when `quarkus.oidc.authentication.par.enabled` is set, the user is authorized access secured endpoint.
```
jcarranzan, line 35: I think this will pass identically with or without PAR, maybe consider replacing with more PAR-specific (redirect url format wh PAR enabled or just should content client_id and request_id)

jcarranzan (changes requested):
I left a couple of comments that need to be taken a look at.
On the other hand, looks good, thanks.

QUARKUS-7174.md
jedla97: Good point, I will check if it can be somehow change on Keycloak side and update it. If not possible I think we can at least check if this property is take into account setting it to something like `http://loclahost:8080/my/par/enpoint/`

QUARKUS-7174.md
jedla97: > Just our of curiosity. What will happen if PAR fails?

For if the authorization server not support it I need to check. I check it only for server require it and Quarkus not setting it. In this case I think it throw 401, but maybe it was different 4xx number.

I will check it today.

QUARKUS-7174.md
jedla97: I plan to set the Keycloak to require the PAR. But tbh I don't fully get PAR-specific exmeple you wrote. If I understand it correctly you suggesting check the redirect url, this will be checked it standalone test.

jcarranzan: Here is the code that I mean: https://github.com/quarkusio/quarkus/blob/724a255cb576c77433ae99fa3feef55e619db932/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java#L762

QUARKUS-7174.md
jedla97: Ok I checked it and it's throw the the 500 error, which is in my pov correct for end app. The log contains `io.quarkus.oidc.OIDCException: {"error":"HTTP 404 Not Found"}` which indicate that the `http://localhost:8180/realms/quarkus/protocol/openid-connect/ext/par/request` was not available so the PAR on Keycloak was not enabled. 

I'll think about how "expensive" would be test this as it would need to use another Keycloak

mocenas: I agree that this seems like a correct behaviour.

I don't think we need to have an automated test that app fails if it has broken config/environment :-)

QUARKUS-7174.md
jedla97: I'll add it

QUARKUS-7174.md
jedla97: This is how the Quarkus communicate wit OIDC provider under the hood. We want to tests that the final app is working, and basically if I understand that code correctly, this will be covered by `Check if Quarkus and Keycloak communicate using back channel and Quarkus redirect user to url with identifier only`

Also Keycloak can enforce only accept PAR, otherwise it will throw 401 (not 100% sure now about code ), so this should tell us if something break and PAR is not used

mocenas (approved):
LGTM

QUARKUS-7174.md
jcarranzan: You're right, if Keykloak enforces PAR, a successful login already proves the back-channel flow worked correctly. It makes sense, thanks.

jcarranzan (approved):
LGTM

PR #309, opened 2026-07-01
QUARKUS-7340 - OIDC: Support for custom DPoP nonce providers

Andy-Y1314:
### Links

JIRA: https://redhat.atlassian.net/browse/QUARKUS-7340

Quarkus documentation: https://quarkus.io/guides/security-oidc-bearer-token-authentication#demonstrating-proof-of-possession-dpop

   - Logging
   - Tracing
   - Metrics
   - Security
   - OpenAPI
   - Data sources
   - Frontend
   - Qute

### DIFF
--- a/QUARKUS-7340.md
+++ b/QUARKUS-7340.md
@@ -1,0 +1,53 @@
+# QUARKUS-7340 - OIDC: Support for custom DPoP nonce providers
+
+JIRA: https://redhat.atlassian.net/browse/QUARKUS-7340
+
+Upstream PR: https://github.com/quarkusio/quarkus/pull/52090
+
+This PR adds support for custom DPoP nonce providers in Quarkus OIDC by introducing the `DPoPNonceProvider` interface. Applications can implement this interface to provide and validate DPoP nonces during authentication.
+
+## Scope of the testing
+
+Tests will cover basic custom DPoP nonce provider behaviour for Quarkus OIDC protected resources.
+
+* A DPoP request with a valid nonce is accepted successfully.
+* A DPoP request without a nonce returns `401 Unauthorized` with the expected DPoP challenge headers (`WWW-Authenticate`, `DPoP-Nonce` and `Cache-control: no-store`).
+* A DPoP request with an invalid nonce returns `401 Unauthorized` with the expected DPoP challenge headers.
+* A client can successfully retry the request using the nonce returned in the initial challenge.
+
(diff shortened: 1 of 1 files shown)


QUARKUS-7340.md
```diff
@@ -0,0 +1,59 @@
...
+
+* A DPoP request with an invalid nonce returns `401 Unauthorized` with the expected DPoP challenge headers.
+
+* A client can successfully retry the request using the nonce returned in the initial challenge.
+
+## Existing test coverage
+
+Existing upstream tests are located here: https://github.com/quarkusio/quarkus/tree/main/integration-tests/oidc-dpop
+
+Upstream test coverage includes tests for the custom `DPoPNonceProvider` flow, including scenarios for missing nonce, invalid nonce and successful retry with a valid nonce.
```
fedinskiy, line 24: What is the difference between the existing scenarios upstream and the ones you plan to add?

QUARKUS-7340.md
```diff
@@ -0,0 +1,59 @@
...
+* Reviewed documentation on DPoP.
+
+* Reviewed this issue https://github.com/quarkusio/quarkus/issues/54854.
+
+## Contacts
+
+* Tester: Andy Yan <andy.yan@ibm.com>
+
+## References
+* https://quarkus.io/guides/security-oidc-bearer-token-authentication#demonstrating-proof-of-possession-dpop
```
fedinskiy, line 52: Nitpick, but half of lists in this file uses has empty lines between items, and the other half hasn't. Could you unify the style? Lists without spacing would be easier to read I presume.

QUARKUS-7340.md
Andy-Y1314: I was thinking because we test more/different environments, for example OpenShift.

With this feature the main combinations are has nonce, no nonce, invalid nonce which I understand is covered in upstream, however what we have currently in the test suite only tests DPoP but with no nonce, so we will be adding coverage for that as well.

QUARKUS-7340.md
Andy-Y1314: I think the style is consistent because I used spacing between every list items except for the reference section... 

I'll remove the empty lines between list items to make it easier to read.

jedla97 (commented):
I think it's look fine, just for info, the coverage contained problem which was not fully with the spec/rfc for it. The content of `wwwAuthenticate` should contain `use_dpop_nonce` instead of `invalid_dpop_proof`. This is fixed by https://github.com/quarkusio/quarkus/issues/54854 and it's planed to be backported to 3.x. I think you should check that the `wwwAuthenticate`  content is correct, but also check if it's not contain invalid_dpop_proof

jedla97 (approved):
Thanks!

fedinskiy (approved):

PR #310, opened 2026-07-15
QUARKUS-6051 - JSON option support for named logging handlers

Andy-Y1314:
### Links

JIRA: https://redhat.atlassian.net/browse/QUARKUS-6051

Quarkus documentation: https://quarkus.io/guides/logging#console-log-handler

   - Logging
   - Tracing
   - Metrics
   - Security
   - OpenAPI
   - Data sources
   - Frontend
   - Qute

### DIFF
--- a/QUARKUS-6051.md
+++ b/QUARKUS-6051.md
@@ -1,0 +1,56 @@
+# QUARKUS-6051 - JSON option support for named logging handlers
+
+JIRA: https://redhat.atlassian.net/browse/QUARKUS-6051
+
+Upstream PR: https://github.com/quarkusio/quarkus/pull/53258
+
+This feature adds support for configuring JSON logging independently for named logging handlers. Previously, if users wanted a named logging handler to produce JSON output, it had to be enabled globally and the named handlers could not override the global JSON logging configuration. With this new feature the named handlers can explicitly enable or disable JSON formatting regardless of the global setting.
+
+## Scope of the testing
+Verify that JSON logging can be configured independently for named logging handlers.
+
+For each named handler type — **console**, **file**, **syslog**, and **socket** — verify both override scenarios:
+
+* **Global JSON off, handler JSON on:** the handler produces JSON output.
+* **Global JSON on, handler JSON off:** the handler produces plain text output.
+
+Additional verification per handler:
+* **File handler:** the configured log file is created in both cases.
+* **Syslog handler:** output is sent to the configured syslog receiver.
+* **Socket handler:** output is sent to the configured TCP receiver.
+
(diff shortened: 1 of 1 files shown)


jcarranzan (commented):
At first glance looks good, I will review a bit more tomorrow.
However, afaik quarkus-logging-json is full support , could u confirm it and change it in the summary description?

jedla97 (approved):
It's look fine to me!

>  didn't add Socket and Syslog coverage to the test plan yet, since upstream tests verify formatter selection for these handler types, while in our case we need to setup additional infrastructure.
I'm wondering if console and file coverage is sufficient or should I also cover socket and syslog WDYT?

Afaik looking and we testing syslog, so it should be ready (don't know how much as I never touch this) see
https://github.com/quarkus-qe/quarkus-test-suite/blob/main/logging/thirdparty/src/test/java/io/quarkus/ts/logging/jboss/SyslogIT.java#L16

For question if you cover this,. I don't have strong opinion here, but if you could look at `SyslogIT` and try to experiment with it to see how hard it would be. If it would be easy as the coverage for console/file I would say add it. If it would be hard like setting additional things and adding complexity I would say it's not worth the hassle.

jcarranzan (approved):
Apart from my small comment, TP looks good to me thanks @Andy-Y1314

QUARKUS-6051.md
```diff
@@ -0,0 +1,55 @@
...
+## Existing test coverage
+Upstream tests for named console, file, socket and syslog handlers by checking the correct formatter is assigned.
+
+The QE coverage will test the feature from a user perspective. The tests will start a Quarkus application, trigger logging through a REST endpoint and verify that the named handler produces the expected JSON or plain text.
+
+The QE test suite contains tests for logging, but does not currently verify JSON configuration for named logging handlers.
+
+### Impact on test suites and testing automation
+
+* The new tests will be added under the `logging/jboss` module.
```
jedla97: Wouldn't make sense to have syslog, maybe even socket tests, in `logging/third-party` module?

jedla97 (commented):
I have just one comment to the new changes. 

Also @Andy-Y1314 Jose have PTO so he probably won't look at it, but I think we can go with this, as most of the TP wasn't been changed.

QUARKUS-6051.md
Andy-Y1314: Yes since there are existing syslog tests there and since sockets use that setup as well, I'll add it there.

jedla97 (approved):
It's look fine to me and compared to the changes which Jose approved, there is no much changes so I think it's ok to merge.

PR #320, opened 2026-08-19
QUARKUS-7819 - Add CORS support to the management interface

Andy-Y1314:
### Links

JIRA: https://redhat.atlassian.net/browse/QUARKUS-7819

Quarkus documentation: https://quarkus.io/guides/management-interface-reference

   - Logging
   - Tracing
   - Metrics
   - Security
   - OpenAPI
   - Data sources
   - Frontend
   - Qute

### DIFF
--- a/QUARKUS-7819.md
+++ b/QUARKUS-7819.md
@@ -1,0 +1,52 @@
+# QUARKUS-7819 - Add CORS support to the management interface
+
+JIRA: https://redhat.atlassian.net/browse/QUARKUS-7819
+
+Upstream PR: https://github.com/quarkusio/quarkus/pull/53432
+
+This PR adds dedicated CORS configuration for the Quarkus management interface using the quarkus.management.cors.* properties.
+
+Management CORS is disabled by default, to make your Quarkus management interface accessible to another application running on a different domain, CORS must now be explicitly configured. Also when CORS is enabled without configuring specific origins, only same origin requests are allowed.
+
+## Scope of the testing
+
+* Verify that a configured origin can access management endpoints such as /q/health.
+* Verify that an origin not included in the management CORS configuration is rejected with HTTP 403.
+* Verify that the same origin requests to the management interface are allowed.
+* Verify that the application and management interfaces can use different CORS policies.
+* Verify that when management CORS is disabled, request to /q/health succeeds but the response does not contain an Access-Control-Allow-Origin header.
+
+## Out of Scope
+* Testing every management endpoint, /q/health will be used as the representative.
+* Running a real client application or browser, requests are simulated like other REST tests.
+
+## Existing test coverage
+
+Upstream tests cover allowed, rejected and same origin requests to the management interface.
+
+The QE test suite does not currently cover management CORS. The new tests will cover upstream scenarios and also verify that application and management interfaces can be configured with different allowed origins.
+
+### Impact on test suites and testing automation
+
+* The new tests will be placed in the `http/management` module.
+
+### Impact on resources
+
+The added tests should have **low impact** on resources:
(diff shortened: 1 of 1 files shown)


QUARKUS-7819.md
```diff
@@ -0,0 +1,45 @@
...
+
+JIRA: https://redhat.atlassian.net/browse/QUARKUS-7819
+
+Upstream PR: https://github.com/quarkusio/quarkus/pull/53432
+
+This PR adds dedicated CORS configuration for the Quarkus management interface using the quarkus.management.cors.* properties.
+
+## Scope of the testing
+
+* Verify that a configured origin can access management endpoints such as /q/health.
```
fedinskiy, line 13: Here and below,  "origin" means an app?

Andy-Y1314: Yes, basically it's the address where the client app is running such as http://localhost:3000, CORS uses the request Origin header to determine whether the application can access the management endpoint.

Andy-Y1314: For example, to allow requests from a specific origin:

quarkus.management.enabled=true
quarkus.management.cors.enabled=true
quarkus.management.cors.origins=http://example.com

Taken from the doc, for example if the client app is running at http://example.com it can then access management endpoints.

fedinskiy: How do we enforce that for apps in the TS? Does that property allow IPs+ports?

Andy-Y1314: Oh we can just simulate it as we do for other REST tests, I did some testing and basically something like this 

`static final RestService app = new RestService()
.withProperty("quarkus.management.cors.enabled", "true")
.withProperty("quarkus.management.cors.origins", "http://localhost:3001");`

`app.management()
                .header("Origin", "http://localhost:3001")
                .get("q/health")
                .then()
                .statusCode(HttpStatus.SC_OK)
                .header("Access-Control-Allow-Origin", is("http://localhost:3001"));`

Obviously not the full code just snippets.

fedinskiy (approved):

QUARKUS-7819.md
```diff
@@ -0,0 +1,45 @@
+# QUARKUS-7819 - Add CORS support to the management interface
+
+JIRA: https://redhat.atlassian.net/browse/QUARKUS-7819
+
+Upstream PR: https://github.com/quarkusio/quarkus/pull/53432
+
+This PR adds dedicated CORS configuration for the Quarkus management interface using the quarkus.management.cors.* properties.
```
jcarranzan, line 7: Ok, but I will be more explicit, the upstream PR is labeled release/breaking-change and triage/security. Now CORS is disabled by default, so I think the TP should mention this behavioral change and include a scenario verifying the default behavior (CORS not enabled, cross-origin requests get no CORS headers...).

jcarranzan: And also, as a feature summary explain the significance of the change and/or its constraints...

QUARKUS-7819.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Scope of the testing
+
+* Verify that a configured origin can access management endpoints such as /q/health.
+* Verify that an origin not included in the management CORS configuration is rejected with HTTP 403.
+* Verify that the same origin requests to the management interface are allowed.
+* Verify that the application and management interfaces can use different CORS policies.
+
```
jcarranzan, line 22: Also, think about what is out of scope and won't be tested

jcarranzan (changes requested):
Thanks Andy, I left a couple of comments, could you review it?

jcarranzan (approved):

